PR #17, opened 2020-02-21
Fixes to Transition Systems

makaimann:
Based on earlier discussion, this PR swaps the inheritance relationship between `RelationalTransitionSystem` and `FunctionalTransitionSystem`. Furthermore, it drops `Relational` and just calls it a general `TransitionSystem`. Furthermore, it enforces that INIT/INVAR/PROP use only state variables. It adds a `constrain_inputs` function which will add a constraint to TRANS only over current states and inputs. There's also a convenience function that picks between `add_invar` and `constrain_inputs`, based on whether the constraint contains inputs.

Because there will sometimes be inputs in the property, the `btor2_encoder` will make a witness for the bad state if necessary. I'm not sure if it makes sense to do this here, or later. But, this was the easiest place to add it now. Note that this can add lots of states to the system: e.g. if the BTOR2 file has lots of bad states that have inputs in them, there's one witness for each of those, even if we're only checking one property. I'm open to other solutions if people have suggestions. In the future, maybe we should copy the transition systems for each property and modify it per property. This also make sense for COI @lonsing .

Note, I'm currently running it on the HWMCC19 benchmarks to check for any potential bugs I introduced.


ahmed-irfan:
@makaimann you mentioned that you were running the HWMCC benchmarks on the cluster. Did the run finish?

makaimann:
26Feb2020-cosa2-runs

Sorry for the delay! Here's a screenshot of the compare script showing that there's no discrepancies.

makaimann:
And here are screenshots showing totals. We lost a couple benchmarks for k-induction. I can look at the code changes to see if there's data structure copies or something:

26Feb2020-cosa2-bmc-totals

26Feb2020-cosa2-ind-totals

26Feb2020-cosa2-interp-totals

makaimann:
I just realized that the unroller needs to be updated so that it works with `FunctionalTransitionSystem`. I don't currently have any tests for that, but it calls `next` on the transition system which is disabled for functional systems. I'll handle that soon.

core/unroller.cpp
```diff
@@ -81,7 +81,7 @@ UnorderedTermMap & Unroller::time_cache_at_time(unsigned int k)
     const unsigned int t = time_cache_.size() - 1;
 
     for (auto v : ts_.states()) {
-      Term vn = ts_.next(v);
+      Term vn = ts_.next_map_.at(v);
```
ahmed-irfan: let's also add a comment about why we are doing this.

makaimann:
I made some more changes, including adding [googletest](https://github.com/google/googletest) infrastructure. I'm running more experiments on the cluster now.

makaimann:
Here are the latest results! No discrepancies with master, the two errors in master are now timeouts, and we only lose 3 benchmarks for `ind` and otherwise we solve the same number of benchmarks. As far as I'm concerned, this now ready, but I'd appreciate opinions on the new TransitionSystem inheritance structure (an abstract parent class, and two implementations -- relational and functional).

Screen Shot 2020-02-28 at 12 26 16 PM

Screen Shot 2020-02-28 at 12 28 12 PM

Screen Shot 2020-02-28 at 12 31 22 PM

CMakeLists.txt
```diff
@@ -62,6 +64,12 @@ if (WITH_MSAT)
   endif()
 endif()
 
+if (WITH_CVC4)
+  if (NOT EXISTS "${PROJECT_SOURCE_DIR}/deps/smt-switch/local/lib/libsmt-switch-cvc4.${SHARED_LIBRARY_EXT}")
+    message(FATAL_ERROR "Missing smt-switch mathsat library -- try running ./contrib/setup-smt-switch.sh --with-cvc4")
```
ahmed-irfan: typo: mathsat -> cvc4

core/ts.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+void TransitionSystem::assign_next(const Term & state, const Term & val)
+{
+  // TODO: only do this check in debug mode
+  if (states_.find(state) == states_.end()) {
+    throw CosaException("Unknown state variable");
+  }
+
+  if (!no_next(val))
```
ahmed-irfan: clang-format

core/ts.cpp
```diff
@@ -0,0 +1,237 @@
...
+  {
+    throw CosaException("Got next state variable in RHS of functional assignment");
+  }
+
+  state_updates_[state] = val;
+  trans_ = solver_->make_term(
```
ahmed-irfan: clang-format

core/ts.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+void TransitionSystem::add_invar(const Term & constraint)
+{
+  // TODO: only check this in debug mode
+  if (only_curr(constraint)) {
+    init_ = solver_->make_term(And, init_, constraint);
+    trans_ = solver_->make_term(And, trans_, constraint);
+    // add the next-state version
+    trans_ = solver_->make_term(
```
ahmed-irfan: clang-format

core/ts.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+bool TransitionSystem::is_next_var(const Term & sv) const
+{
+  return (next_states_.find(sv) != next_states_.end());
+}
+
+// protected methods
+
+bool TransitionSystem::only_curr(const Term & term) const
```
ahmed-irfan: maybe we can create a common function contains(const Term & term, TermVec vec). 
only_curr(term t) := contains(t, states_)
no_next(term t) := contains(t, states_ + inputs_)

cosa2.cpp
```diff
@@ -203,7 +204,7 @@ int main(int argc, char ** argv)
       s->set_opt("incremental", "true");
     }
 
-    RelationalTransitionSystem rts(s);
+    FunctionalTransitionSystem rts(s);
```
ahmed-irfan: rename rts to fts

utils/term_analysis.cpp
```diff
@@ -0,0 +1,58 @@
...
+** Copyright (c) 2019 by the authors listed in the file AUTHORS
+** in the top-level source directory) and their institutional affiliations.
+** All rights reserved.  See the file LICENSE in the top-level source
+** directory for licensing information.\endverbatim
+**
+** \brief Useful functions for term analysis.
+**
+**
+**/
+
```
ahmed-irfan: clang-format

utils/term_analysis.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "smt-switch/smt.h"
+
+using namespace smt;
+
+namespace cosa
+{
+
+void get_free_symbols(Term & term, UnorderedTermSet & out_symbols)
+{
+  smt::TermVec to_visit({term});
```
ahmed-irfan: no need for smt::

utils/term_analysis.cpp
```diff
@@ -0,0 +1,58 @@
...
+  smt::UnorderedTermSet visited;
```
ahmed-irfan: ditto

utils/term_analysis.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+  smt::Term t;
```
ahmed-irfan: smt::

utils/term_analysis.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+      if (t->is_symbolic_const())
+      {
+        out_symbols.insert(t);
+      }
+    }
+  }
+}
+
+UnorderedTermSet get_free_symbols(smt::Term & term)
```
ahmed-irfan: smt::

CMakeLists.txt
makaimann: Nice catch! Fixed with f038c80e2cd248cc3a2fb367ed3d53946b14e123

core/ts.cpp
makaimann: Formatted with 1308d7948ff2db5bfc0cdc89437f6581526bf400

core/ts.cpp
makaimann: Formatted with 1308d7948ff2db5bfc0cdc89437f6581526bf400

core/ts.cpp
makaimann: Formatted with 1308d7948ff2db5bfc0cdc89437f6581526bf400

utils/term_analysis.cpp
makaimann: Formatted with 1308d7948ff2db5bfc0cdc89437f6581526bf400

utils/term_analysis.cpp
makaimann: Done with fcd3fd8a27e3ee5718e33db5d3d3341a94edb2c6

utils/term_analysis.cpp
makaimann: Done with fcd3fd8a27e3ee5718e33db5d3d3341a94edb2c6

utils/term_analysis.cpp
makaimann: Done with fcd3fd8a27e3ee5718e33db5d3d3341a94edb2c6

utils/term_analysis.cpp
makaimann: Done with fcd3fd8a27e3ee5718e33db5d3d3341a94edb2c6

core/ts.cpp
makaimann: Good idea! Implemented with b78fe5a29c59031d8c97619c21db3864ef2e1363

makaimann:
Because the change in b78fe5a29c59031d8c97619c21db3864ef2e1363 is somewhat non-trivial, I'm going to rerun the benchmarks to ensure I didn't introduce any bugs. I'll post the results asap.

ahmed-irfan:
updates on the new experiment run...

makaimann:
Here are the latest results, everything looks good:

latest-ts-fixes-comparison
latest-ts-fixes-totals

PR #24, opened 2020-03-12
Add python bindings for cosa2

makaimann:
Uses cython to build python bindings for cosa2. See README.md for installation instructions.


README.md
```diff
@@ -26,3 +26,23 @@ For more information, see the examples in `smt-switch/tests/btor`. Other useful
...
+To build the `cosa2` python bindings, first make sure that you have [Cython](https://cython.org/) version >= 0.29 installed. Then ensure that `smt-switch` and its python bindings are installed. Finally, you can configure with `./configure.sh --python` and then build normally. The sequence of commands would be as follows:
+```
+# Optional recommended step: start a python virtualenv
+# If you install in the virtualenv, you will need to activate it each time before using pycosa2
+# and deactivate the virtualenv with: deactivate
+python3 -m venv env
+source ./env/bin/activate
+pip install Cython==0.29 pytest
+./contrib/setup-smt-switch.sh --python
+pip install -e ./deps/smt-switch/build/python
```
lonsing: A minor glitch in the steps:
`./contrib/setup-btor2tools.sh`
must be run before
`./configure.sh --python`

makaimann: Good catch, thanks! Updated with: d7350f544ac067ba956aee1e745fe8806ba7c9a2

configure.sh
```diff
@@ -17,6 +17,8 @@ Configures the CMAKE build environment.
                         Required for interpolant based model checking
 --with-cvc4             build with CVC4 support (default: off)
 --debug                 build debug with debug symbols (default: off)
+--python                compile with python bindings (default: off)
+--py2                   use python2 interpreter (default: python3)
```
ahmed-irfan, line 21: python2 reached its end-of-life. Should we support it?

utils/logger.cpp
```diff
@@ -20,4 +20,6 @@
 // declare a global logger
 namespace cosa {
 Log logger;
+
+void set_global_logger_verbosity(unsigned int v) { logger.set_verbosity(v); }
```
ahmed-irfan, line 24: clang format

makaimann: Ran it again, looks like that's what clang-format does actually.

configure.sh
makaimann: Hmm that's a good question... I wouldn't be surprised if there are still a couple people using it, but maybe it's not worth the headache. I don't have a specific use case in mind, except that I know Mac can sometimes have issues with python3 bindings (issues linking the right python library).

utils/logger.cpp
ahmed-irfan: ok :)

makaimann:
Yeah it’s a good point. I mentioned above that there still might be some people using it. I also know that Mac can have issues linking python3 correctly for bindings. This can be worked around but some people might just decide to use python 2. It’s not a great use case but it’s something. What do you think?

ahmed-irfan:
Ok. I think we can keep the Python2 support for now.

makaimann:
I don't know why Travis is failing on the PR -- the commits in the branch are passing. I'm going to try merging and can revert if master fails.

PR #30, opened 2020-04-13
WIP: Imp unroller

ahmed-irfan:


makaimann:
It's looking good to me so far. Do you think the untime cache is a bottleneck? What if we just don't have it all, we could instead just substitute in reverse.

ahmed-irfan:
untime_cache was not the bottleneck. I just made it more lightweight. 
On the other hand, time_var_cache_ was the bottleneck. So, I removed it because we could live without it.

ahmed-irfan:
this is buggy

PR #32, opened 2020-04-22
Switch back to badvec in btor-encoder

makaimann:
I recently noticed that we get an incorrect result (actually the result is the same but the bound is wrong). The bug in the attached btor2 file should be triggered around bound 36 or something, but instead it's triggered at bound 7. Perhaps even worse, it was nondeterministic -- sometimes it would fail at bound 7, and changing something unrelated like the verbosity would make it get to bound 10 without a problem.

By going through the cosa2 history I found that changes to the btor2 encoder caused this. On my machine, this works. However, I have no idea why it's any different. It should be functionally equivalent. I'm curious what the performance is on another machine.

[btor-nondet.tar.gz](https://github.com/upscale-project/cosa2/files/4513177/btor-nondet.tar.gz)

### DIFF
--- a/cosa2.cpp
+++ b/cosa2.cpp
@@ -258,16 +258,16 @@ int main(int argc, char ** argv)
       logger.log(2, "Parsing BTOR2 file: {}", filename);
       FunctionalTransitionSystem fts(s);
       BTOR2Encoder btor_enc(filename, fts);
-      const TermVec & propvec = btor_enc.propvec();
-      unsigned int num_props = propvec.size();
+      const TermVec & badvec = btor_enc.badvec();
+      unsigned int num_props = badvec.size();
       if (prop_idx >= num_props) {
         throw CosaException(
             "Property index " + to_string(prop_idx)
             + " is greater than the number of properties in file " + filename
             + " (" + to_string(num_props) + ")");
       }
-      Term prop = propvec[prop_idx];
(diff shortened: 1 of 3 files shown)


lonsing:
I tried it out with fresh builds of the 'master' (f930274dc0915a506be83f13f13f2fe29b196add) and 'nondet-bugfix' (1f02922546d31f52e8188baacac307a71d5ab55f) branches, respectively. 

I have not yet tested the 'nondet-bugfix' branch up to bound 36 but only with smaller bounds.

On 'master' I was not able to reproduce nondeterministic behavior. With default parameters ('./cosa2 btor-nondet.btor'), the result is always 'sat' at bound 7. 

However, when adding e.g. '-e bmc -k 8' then the result is unknown (deterministically, as it seems) at the specified maximum bound of 8. 

And when adding '-v 1 -e bmc -k 8' then again the result is always sat at bound 7.

makaimann:
Thanks @lonsing! I shouldn’t have said nondeterministic actually, just that it changes with seemingly unrelated things like the options. It’s interesting that this change fixes it for you as well. This is very strange. I’d like to understand this better. I’m wondering if it has something to do with negation in the smt-switch boolector backend. I’ll look into it more.

makaimann:
We're continuing to debug this issue -- but there's no reason to merge this change.

PR #33, opened 2020-04-29
CoreIR Encoder

makaimann:
First pass on CoreIR encoder. It only creates variables for states and global inputs, thus reducing the load on the solver. It accomplishes this by traversing the circuit starting from register outputs and nodes that have no inputs and finally resulting in the input to registers. It currently only supports circuits where the clock can be abstracted (single clock with no asynchronous behavior). It should be a straightforward extension to add an explicit clock representation for multiple clocks and/or asynchronous resets.


makaimann:
@rdaly525 If you have a chance to look at this, I'd appreciate it. I think you can add a review, but strangely I don't know if I can request a review from you since you're not in this organization. By the way, I'm waiting a bit to merge some PRs so this is currently on another branch -- but once I merge `python` I'll switch this PR to merge to master.

frontends/coreir_encoder.cpp
```diff
@@ -0,0 +1,331 @@
...
+using namespace CoreIR;
+using namespace smt;
+using namespace std;
+
+// helpers
+
+bool instance_of(CoreIR::Instance * inst, std::string ns, std::string name)
+{
+  auto mod = inst->getModuleRef();
+  if (!mod->isGenerated()) {
```
rdaly525, line 17: Any module from corebit will not be generated, so this function will not work as intended.

makaimann: Interesting, so is the right thing to do not to check `isGenerated` at all?

makaimann:
Great, thanks for the review! I've only run it on `counters.json` from the `coreir` examples so far. I'll add some tests soon. Is there a good place to get a larger set of coreir files?

frontends/coreir_encoder.cpp
rdaly525: Well, it is a bit annoying due to the fact that "coreir" (aka BitVector) has a generator parameter of "width" whereas "corebit" (aka Bool) does not have any generator parameters. So the easiest for this code is to just have a function that checks if an instance is a primitive. This can be accomplished by just checking if the namespace associated with the module is either "coreir" or "corebit".

PR #39, opened 2020-05-25
Scanner fixes

yangy96:


makaimann:
All right, I ended up just manually including the brew install location for the flex header files if it's being built on Mac. I'm hoping to have a better solution soon, but this works for now.

frontends/smv_encoder.cpp
```diff
@@ -1,2 +1,60 @@
...
+        [](smt::SmtSolver solver_) {
+          Result r = solver_->check_sat();
+          return r.is_sat();
+        },
+        solver_);
+    std::future_status status;
+    status = fut.wait_for(std::chrono::seconds(5));
+    while (status != std::future_status::timeout) {
+      if (status == std::future_status::ready) {
+        if (fut.get() == false) {
```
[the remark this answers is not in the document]
makaimann: And actually, I think my preference would be to have the positive case in the if statement. So you can throw the exception in the if and add to trans in the else. Or, you could instead check `is_unsat` and keep the order the same but just check `fut.get()`. I think the latter is probably easier to understand actually.

frontends/smv_encoder.cpp
```diff
@@ -1,2 +1,60 @@
...
+          rts_.constrain_trans(casestore_[element.first]);
+          break;
+        } else {
+          throw CosaException("case error");
+        }
+      }
+    }
+    if (status == std::future_status::timeout)
+      throw CosaException("case timeout check error");
```
makaimann, line 57: This is just a style preference of mine: I'd prefer to have the brackets around this statement, even though it's functionally correct as is.

frontends/smvparser.y
```diff
@@ -446,14 +564,23 @@ basic_expr: constant {
...
+                    $$ = new SMVnode(res);
+                  } else if (bvs_a == bvs_b == SMVnode::Signed){
+                    smt::Term res = enc.solver_->make_term(smt::BVSge, a->getTerm(), b->getTerm());
+                    $$ = new SMVnode(res);
+                  } else{
+                    throw CosaException ("Unsigned/Signed unmatch");
+                  } 
               }
             }
             | OP_MINUS basic_expr %prec UMINUS{
```
makaimann, line 580: This is just for me to understand. What is the `%prec` here?

frontends/smvparser.y
```diff
@@ -487,8 +626,14 @@ basic_expr: constant {
...
+                  if(bvs_a != bvs_b){
+                   throw CosaException("Unsigned/Signed unmatch");
+                  } else{
                   smt::Term res = enc.solver_->make_term(smt::BVMul, a->getTerm(), b->getTerm());
-                  $$ = new SMVnode(res);
+                  $$ = new SMVnode(res,bvs_a);
+                  } 
               }
             }
             | basic_expr "/" basic_expr{
```
makaimann, line 639: Is there a reason this op isn't called `OP_DIV`?

frontends/smvparser.y
```diff
@@ -511,33 +665,62 @@ basic_expr: constant {
...
+                throw CosaException("Unsigned/Signed unmatch");
+              } else{
+                smt::Term res = enc.solver_->make_term(smt::Concat, a->getTerm(), b->getTerm());
+                $$ = new SMVnode(res,SMVnode::Unsigned);
+              }               
+            }
+            | basic_expr "[" integer_val "]" {
+                throw CosaException("No index Subscript");
             }
             | basic_expr "[" basic_expr ":" basic_expr "]"{
```
makaimann, line 711: It might be better to be more specific here (if that's not too difficult). The extract operator has to take non-negative integers (just saying integers would be fine). So instead of `basic_expr`, I think it could be `integer_constant`?

frontends/smvparser.y
```diff
@@ -511,33 +665,62 @@ basic_expr: constant {
...
-              throw CosaException("No word1");
+                SMVnode *a = $1;
+                SMVnode *b = $3;
+                SMVnode *c = $5;
+                SMVnode::BVtype bvs_a = a->getBVType();
+                SMVnode::BVtype bvs_b = b->getBVType();
```
makaimann, line 716: Why is this `getBVType()` when we would expect `b` and `c` to be integers?

frontends/smvparser.y
```diff
@@ -250,7 +279,7 @@ constant: boolean_constant {
           | integer_constant {
```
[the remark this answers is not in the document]
makaimann: Also, what about boolean symbols? I.e. a `VAR` that's type `boolean`?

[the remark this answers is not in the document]
yangy96: Hi Makai,

I think I supported boolean value TRUE and FALSE using boolean constant in line 348. I can add another type SMVnode named boolean. Does it work better?

frontends/smvparser.y
yangy96: I used symbols instead of OP_ADD, OP_MUL for all arithmetic operations. Do you prefer OP_DIV?

frontends/smvparser.y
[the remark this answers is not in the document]
makaimann: Oh I see — I missed it because it was collapsed since it wasn’t involved in the diff. I think it’s fine how it is. Except, how do you handle VARs of type Boolean without a Boolean SMVNode?

frontends/smvparser.y
makaimann: Oh no that’s okay — as long as there’s a pattern. Sounds good!

frontends/smvparser.y
yangy96: I followed bison's manual to use %prec for unary minus in the grammar

"The %prec modifier declares the precedence of a particular rule by specifying a terminal symbol whose precedence should be used for that rule."

frontends/smvparser.y
yangy96: I want to do type checking here, so the type of the first expression should be word

frontends/smvparser.y
yangy96: I think in the manual, it says the type of basic_expr should be an array? Do you want me to change `basic_expr` to `array`?

frontends/smvparser.y
makaimann: Thanks!

frontends/smvparser.y
makaimann: No I wouldn't -- there's a bit of a terminology mismatch here. What they're calling an array, we would call a bit-vector (an array of bits). An array in our terminology is an SMT array -- it's a data structure that can be defined over arbitrary index and element sorts (e.g. an array from integers to bit-vectors). So what I would actually expect to see is:
```
| bitvec_expr "[" integer_constant ":" integer_constant "]"
```

but you don't need to actually have a dedicated bit vector expression. I think it's probably fine as-is, but I'm curious how it fails if you did something like this:

```
VAR b : unsigned word[8];
VAR t : boolean;
INIT b[t:t] = 0ud1_1
```

because `b[t:t]` is not supported, it would have to be integer values like `b[0:0]`

makaimann: actually I see that you check they're integers below.

frontends/smvparser.y
makaimann: Hmm not quite. So I agree that the first expression should be a word, so that makes sense that it's type BV. But the second two expressions should be integers, not bit-vectors. So why is it `SMVNode::BVtype bv_b = b->getBVType();` I see that you check if it's an integer below, but I'm confused why it's called a `BVType` in the first place. It seems like it should just be a "Type" and then you check whether it's an integer or not. Do you know what I mean?

frontends/smvparser.y
```diff
@@ -511,33 +665,62 @@ basic_expr: constant {
...
             | basic_expr "[" basic_expr ":" basic_expr "]"{
-              throw CosaException("No word1");
+                SMVnode *a = $1;
+                SMVnode *b = $3;
+                SMVnode *c = $5;
+                SMVnode::BVtype bvs_a = a->getBVType();
+                SMVnode::BVtype bvs_b = b->getBVType();
+                SMVnode::BVtype bvs_c = c->getBVType();
+                if(bvs_a == SMVnode::BVnot || bvs_b != SMVnode::Integer || bvs_c != SMVnode::Integer){
+                  smt::Term res = enc.solver_->make_term(smt::Extract, a->getTerm(), b->getTerm(),c->getTerm());
```
makaimann, line 717: Also, this isn't the right way to apply `Extract` actually. `Extract` is an indexed operator, so you'd do it like this:
```
enc.solver_->make_term(smt::Op(smt::Extract, 
                                                          <b as a C++ integer>, 
                                                          <c as a C++ integer>), 
                                           a->getTerm());
```

Would it be possible to add a test for this? I think what might help clarify some of the confusion on this part is how Extract can be used e.g. the `[high_bit:low_bit]` syntax in nuXmv. The `high_bit` and `low_bit` cannot be variables, they have to be integer values, e.g. `[2:0]` is fine but not `[a:b]`. So you shouldn't be using `b` and `c` as terms, they should be C++ integers.

frontends/smv_node.h
```diff
@@ -19,20 +18,46 @@ struct SMVnode
...
   };
+  enum Type
+  {
+    Signed,
+    Unsigned,
+    Integer,
+    Real,
+    Boolean,
+    BV,
+    BVnot
```
makaimann, line 29: I remember you mentioning this, but could you explain it again for me please? So the BV tag is for an edge case in Boolector right? Because otherwise I would expect it to just be Signed/Unsigned. And then, what is BVnot for? Is that like the bit-vector operator bitwise NOT?

makaimann:
Merging this now. Additional improvements will be added in https://github.com/upscale-project/pono/pull/48

PR #40, opened 2020-05-28
Support an install target through cmake

makaimann:
Adds an install target for cosa2. It also changes the header includes so that they work when compiling a separate program that depends on cosa2.


frontends/smv_encoder.h
```diff
@@ -6,8 +6,8 @@
 #include <string>
 #include <unordered_map>
 #include "assert.h"
-#include "exceptions.h"
-#include "rts.h"
+#include "utils/exceptions.h"
+#include "core/rts.h"
 #include "smt-switch/smt.h"
 #include "smvparser.h"
```
ahmed-irfan, line 12: do you need to also add frontends to the smvparser.h?

makaimann: Actually I don't think we do, but this isn't working quite right yet. smvparser.h is a generated file, so it needs to be installed correctly. I'll leave more info in another comment.

makaimann:
Just FYI, I'm tabling this PR until https://github.com/upscale-project/cosa2/pull/39 is merged. We should install the generated files for the SMV parser as well, and there are some changes in that PR that should be merged first.

frontends/smv_encoder.h
makaimann: Okay, I updated this now.

makaimann:
All right, I updated this branch. I'm still not totally happy with how things are installed. In particular, I haven't decided if it makes sense to have them all kept in different folders (as they are now) or to group them all together in one directory. Regardless, this works for now and I think it would be good to merge this and we can always change things later. Could I just get a confirmation from @lonsing  or @ahmed-irfan  that it still looks good to merge? Thanks!

lonsing:
Thanks, @makaimann, looks good to me!

PR #41, opened 2020-06-03
Only keep variables as keys in the unroller cache

makaimann:
All right, so @lonsing and I, with lots of help from the boolector developers have been looking into https://github.com/upscale-project/cosa2/issues/36. We're pretty sure that it happens because the boolector `nodemap` which is used in `smt-switch` for the `substitute` function only supports variable -> term substitutions. Because we use the cache in `substitute` in the unroller, this could add non-variables as keys which could cause problems. I added an assertion checking this in `smt-switch` and then made this change so that the cache is always over variables.

Note, one alternative is to not use `nodemap` for substitution, instead relying on the `smt-switch` implementation of `substitute`. However, I would expect that to be much slower because it involves unpacking all the wrapped objects at each subterm. I think the main purpose of the cache was for quick look ups of variables (which still happens), so this fix makes sense to me. What do you guys think?

@lonsing would it be difficult for you to port this change to your setup for reproducing the bug and make sure that it fixes it there also? I checked using the latest versions of everything.

### DIFF
--- a/core/unroller.cpp
+++ b/core/unroller.cpp
@@ -33,6 +33,13 @@ Unroller::~Unroller() {}
 Term Unroller::at_time(const Term & t, unsigned int k)
 {
   UnorderedTermMap & cache = var_cache_at_time(k);
+
+  // if t is a variable, it will be cached
+  auto it = cache.find(t);
+  if (it != cache.end()) {
+    return it->second;
+  }
+
   Term ret = solver_->substitute(t, cache);
   untime_cache_[ret] = t;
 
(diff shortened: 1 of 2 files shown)


ahmed-irfan:
here is the PR I mentioned earlier https://github.com/upscale-project/cosa2/pull/42

ahmed-irfan:
can we run an experiment using the competition benchmarks?

makaimann:
Yeah that's a good idea. I'm still going to merge this now though because this still fixes a bug. But we can run experiments to see how much it affects performance.

makaimann:
It looks like we pick up a few more benchmarks overall -- but it's not a huge difference.

The hwmcc19 "practice" benchmarks:

Screen Shot 2020-06-08 at 11 24 50 AM

Screen Shot 2020-06-08 at 11 26 29 AM

Benchmarks from the POR TACAS paper. Only comparing BMC because these are all false.
Screen Shot 2020-06-08 at 11 27 42 AM

PR #44, opened 2020-06-15
Solve in a new context for each engine

makaimann:
This PR pushes a new context in each engine to keep solver instance clean even when reset_assertions not supported. This also requires computing the witness preemptively, because we cannot get values after popping a context.

The main motivation for this PR is that there are users that plan to interact with the API soon (as a backend for https://github.com/leonardt/fault. I don't want them to have to worry about polluting the solver if they try calls to multiple engines in the same session. Since some solvers don't have `reset_assertions`, the easiest thing to do is to solve in a separate context.

I'm running jobs now to compare performance. If it's not a major performance hit, I think this makes sense. I'll update this PR with results when I have them. In the future, we can think about other alternatives, such as translating terms to a different solver instance first.

### DIFF
--- a/engines/bmc.cpp
+++ b/engines/bmc.cpp
@@ -42,11 +42,16 @@ void Bmc::initialize()
 
 ProverResult Bmc::check_until(int k)
 {
+  // keep the solver instance clean by checking in a different context
+  solver_->push();
   for (int i = 0; i <= k; ++i) {
     if (!step(i)) {
+      compute_witness();
+      solver_->pop();
       return ProverResult::FALSE;
     }
   }
+  solver_->pop();
   return ProverResult::UNKNOWN;
 }
 
--- a/engines/bmc_simplepath.cpp
+++ b/engines/bmc_simplepath.cpp
@@ -32,16 +32,22 @@ BmcSimplePath::~BmcSimplePath() {}
 
 ProverResult BmcSimplePath::check_until(int k)
 {
+  // keep the solver instance clean by checking in a different context
+  solver_->push();
   for (int i = 0; i <= k; ++i) {
     logger.log(1, "Checking Bmc at bound: {}", i);
     if (!base_step(i)) {
+      compute_witness();
+      solver_->pop();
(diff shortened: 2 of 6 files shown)


engines/bmc.cpp
```diff
@@ -42,11 +42,16 @@ void Bmc::initialize()
 
 ProverResult Bmc::check_until(int k)
 {
+  // keep the solver instance clean by checking in a different context
+  solver_->push();
   for (int i = 0; i <= k; ++i) {
     if (!step(i)) {
```
ahmed-irfan, line 48: step asserts unrolled version of the transition relation to the solver. With this change, those assertions will be removed. This is a problem with the current implementation of the step function because it has a check for reached_k (https://github.com/upscale-project/cosa2/blob/49ff98337389ca88a6a478857b5a59704ff7b3de/engines/bmc.cpp#L60).

ahmed-irfan: I think that other engines will have the same problem

ahmed-irfan:
an alternative option is to have a boolean initialized_ (set to false in the constructor) which stores if the initialize() method has already been called. then in the initialize method, we can do something like:
```
if (!initialized_) {
     solver_->push();
     initialized_ = true;
} else {
     solver_->pop();
     solver_->push();
}
```
Now the API user has to make sure that he calls the initialize method before using the particular engine.

engines/bmc.cpp
makaimann: Are you sure? All of the `step` calls are within the same context. `solver_->push()` is called before the calls to `step` and it's not popped until just before a result is returned (after all the `step` calls). That shouldn't be a problem right?

makaimann:
I like that alternative as well. It would probably be easier to maintain.

makaimann:
We discussed this further offline. The main issue is if `check_until` is called multiple times -- which is definitely a reasonable thing for a user to want to do. I'm going to close this PR and open another one soon that translates terms to a fresh solver instance for each engine.

PR #48, opened 2020-06-22
initial change for modular smv

yangy96:


frontends/smv_encoder.cpp
```diff
@@ -1,2 +1,86 @@
...
+    exit(-1);
+  }
+  std::istream & s = ifs;
+  pono::SMVscanner smvscanner(*this);
+  smvscanner.switch_streams(&s);
+  pono::smvparser parse(smvscanner, *this);
+  return parse();
+}
+// parse string
+int pono::SMVEncoder::parseString(std::string newline)
```
makaimann, line 22: If it's not too difficult, it would be really cool if `parseString` could return a `Term` instead of directly adding it to the transition system. That way you can do something like `Term xp1 = smv_encoder.parseString("x + 0ud4_1");`

frontends/smv_node.h
```diff
@@ -10,30 +11,1276 @@
 #include "exceptions.h"
 #include "smt-switch/smt.h"
 
+using namespace std;
 /* node to help construct syntax tree in the parser */
```
makaimann, line 15: Could you expand on this comment a little more after adding the VarType/NodeMtype/Type and element_node? It'd be nice to have some brief explanation here in the code. I think you can just take some bullet points from the slides.

frontends/smv_encoder.cpp
```diff
@@ -1,2 +1,86 @@
...
+}
+
+//modular SMV preprocess
+std::stringstream pono::SMVEncoder::preprocess()
+{
+  module_node * main_n;
+  if (module_list.find("main") != module_list.end()) {
+    main_n = module_list.find("main")->second;
+  } else {
+    throw PonoException("no main module find");
```
makaimann, line 79:
```suggestion
    throw PonoException("no main module found");
```

frontends/smv_node.h
```diff
@@ -10,30 +11,1276 @@
...
+  virtual std::vector<SMVnode *> get_list() { return pa_li; }
+};
+
+class module_node
+{
+  std::string module_name;
+  std::string par_name;
+  std::vector<string> par_li;
+  unordered_map<string, SMVnode *> new_par;
+  element_node * var_li;
```
makaimann, line 140: Just to check my understanding: these `element_node` pointers each store a vector of `SMVNode *` that contain the corresponding type. As in, `var_li` has a vector of `SMVNode *`'s that correspond to variables. Is that right? Could you add a comment about that if so?

frontends/smv_node.h
```diff
@@ -10,30 +11,1276 @@
...
+      if (it->first == SMVnode::FROZENVAR) frozenvar_li = it->second;
+      if (it->first == SMVnode::INIT) init_li = it->second;
+      if (it->first == SMVnode::TRANS) trans_li = it->second;
+      if (it->first == SMVnode::INVAR) invar_li = it->second;
+      if (it->first == SMVnode::INVARSPEC) invarspec_li = it->second;
+    }
+  }
+  unordered_map<string, SMVnode *> get_namelist() { return new_par; }
+  std::string get_par() { return par_name; }
+  void pre(std::unordered_map<std::string, module_node *> module_list,
```
makaimann, line 218: Could you add a comment explaining the difference between `pre` and `preprocess`?

frontends/smvparser.y
```diff
@@ -1187,22 +1194,24 @@ next_expr: TOK_NEXT "(" basic_expr ")"{
...
           for (int i = 0; i < total_num; i++){
-            std::pair<smt::Term,smt::Term> term_pair = enc.caseterm_.back();
+            std::pair<SMVnode*,SMVnode*> term_pair = enc.caseterm_.back();
             enc.caseterm_.pop_back();
             case_true = true;
-            smt::Term e = enc.solver_->make_term(smt::Ite, term_pair.first, term_pair.second, final_term);
-            cond = enc.solver_->make_term(smt::BVOr,cond, term_pair.first);
+            t = term_pair.second->getType();
+            smt::Term e = enc.solver_->make_term(smt::Ite, term_pair.first->getTerm(), term_pair.second->getTerm(), final_term);
+            cond = enc.solver_->make_term(smt::BVOr,cond, term_pair.first->getTerm());
```
makaimann, line 1209: This should be `Or`, not `BVOr`. This might work for boolector, but for all other solvers you can't use a bit-vector operator on booleans. And the conditions should always be boolean.

frontends/smvparser.y
```diff
@@ -1187,22 +1194,24 @@ next_expr: TOK_NEXT "(" basic_expr ")"{
...
             final_term = e;
           }
-          enc.casestore_[case_start] = final_term;
-          enc.casecheck_[case_start] = cond;
-          $$ = new SMVnode(final_term);
+          //enc.casestore_.push_back(final_term);
```
makaimann, line 1212: Remove this if not needed.

frontends/smvparser.y
```diff
@@ -104,6 +104,11 @@ DOT ".";
 header:
     module_decl
     | header module_decl
+    | basic_expr {
+      SMVnode *a = $1; 
+      enc.parse_term = a->getTerm();
+      std::cout << "parse string to term" << enc.parse_term <<std::endl;
```
makaimann, line 110: You can remove this print.

frontends/smv_encoder.h
```diff
@@ -59,9 +61,9 @@ class SMVEncoder
   std::unordered_map<std::string, smt::Term>  unsignedbv_;
   std::deque<std::pair<int, smt::Term>> transterm_;
   std::vector<int> caselist_;
-  std::unordered_map<int, smt::Term> casecheck_;
-  std::unordered_map<int, smt::Term> casestore_;
-  std::vector<std::pair<smt::Term, smt::Term>> caseterm_;
+  std::vector<smt::Term> casecheck_;
```
makaimann, line 64: Could you add short comments for these three vectors please? And actually the more comments the better for these data structures. You can use this comment format: `///< describe what this data structure is used for`. Example:
```
///< vector of booleans, each element is an Or of all the conditions in a case statement.
///< used to check that the conditions cover all possibilities (required by nuXmv manual)
```

frontends/smvparser.y
```diff
@@ -371,46 +575,64 @@ exponential_number: integer_val exponential_prefix "-" integer_val{  $$ =  $1 +
...
             }
             | OP_NOT basic_expr {
+            if(enc.module_flat){
               SMVnode *a = $2;
               SMVnode::Type bvs_a = a->getType();
-              if(bvs_a != SMVnode::BV && bvs_a != SMVnode::Unsigned && bvs_a != SMVnode::Signed && bvs_a != SMVnode::Boolean){
+              if(bvs_a != SMVnode::Unsigned && bvs_a != SMVnode::Signed && bvs_a != SMVnode::Boolean){
                 throw PonoException("Type system violation");
               }
               smt::Term e = enc.solver_->make_term(smt::BVNot, a->getTerm());
```
makaimann, line 620: This should be `Not`, not `BVNot`. I think this is also because it works with Boolector but not with other solvers. You can force the binary to use mathsat by running it with `-e interp`. Everything should still work in that case. I found this by running pono like this from the build directory: `./pono -e interp -v 1 ../samples/simple_counter.smv`

frontends/smvparser.y
```diff
@@ -437,16 +663,20 @@ simple_expr: constant {
                 smt::Term e = enc.solver_->make_term(smt::BVOr, a->getTerm(), b->getTerm());
```
makaimann, line 663: So it looks like all of these operators are using `BV*` ops, when I think they should be boolean ops. Or, you might actually need to check the sorts and use the appropriate operator if they're overloaded (meaning it uses the same symbol in SMV for Or over booleans and bit-vectors). If it is overloaded, then you can do the same thing you did for arithmetic operators, i.e.

```
Term at = a->getTerm();
Term bt = b->getTerm();
SortKind ask = a->get_sort()->get_sort_kind();
SortKind bsk = b->get_sort()->get_sort_kind();
if (at->get_sort() != bt->get_sort() || (ask != BOOL && ask != BV))
{
  throw PonoException("Expecting two booleans or two bit-vectors of the same width");
}

assert(ask == bsk);
assert(ask == BOOL || ask == BV);
Term e;
if (ask == BOOL)
{
   e = enc.solver_->make_temr(smt::Or, at, bt);
}
else
{
   e = enc.solver_->make_term(smt::BVOr, at, bt);
}

assert(e); // check that e is non-null
$$ = new SMVnode(e, bvs_a);
```

makaimann (approved):
Looks good to me. Thanks for the updates! I think this is ready to merge.

PR #49, opened 2020-06-29
Rename to pono

makaimann:
The repository has already been renamed pono. This PR would rename all variations of "Cosa2"/"Cosa" to "Pono" in the codebase.

FYI @zhanghongce, @yangy96, @eshan-singh, @leonardt, @clairexen,
we're renaming cosa2 to "pono".

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,9 +1,9 @@
-The authors, designers, and main contributors to cosa2 are listed
-below. cosa2's copyright is held by these individuals and the
+The authors, designers, and main contributors to pono are listed
+below. pono's copyright is held by these individuals and the
 affiliated institutions at the time of their contributions. See the
-file LICENSE for details on the copyright and licensing of cosa2.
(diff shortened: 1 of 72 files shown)


ahmed-irfan:
just to be sure, are we going to write `Pono` or `pono`?

makaimann:
Haha that's a good question... I'm not picky. I like `pono` for the binary, but if we're referring to the tool it's fine to capitalize. What do you all think?

Also, Travis is failing in the Mac builds now. It looks like it's just failing to set up the environment correctly which happens sometimes. Let's just wait for now and if it fails when I restart it later, I'll look into it more.

lonsing:
>  I like `pono` for the binary, but if we're referring to the tool it's fine to capitalize.

I agree.

ahmed-irfan:
So, it means in the text we should use `Pono` (for example in the AUTHORS file)

PR #51, opened 2020-07-10
Handle command line options in a separate PonoOptions class

lonsing:
Changes in this PR:
- Move the handing of command line options from `main` function in `pono.cpp` to a new class `PonoOptions`. The purpose of this change is to simplify the main function and make option handling more extensible (e.g. as future work, we might want to set options via calls of an API). Also, with this change, options can be accessed from any class by referencing a global `PonoOptions` object. As for the `Logger` object, exactly one `PonoOptions` object exists. 
- Define a new `enum PonoResult` representing the return value of the `main` function in `pono.cpp` (formerly called `int status_code`) to avoid the use of magic constants.


options/options.h
```diff
@@ -0,0 +1,73 @@
...
+        throw PonoException("Cannot create objects of class PonoOptions, "\
+                            "use global object 'pono_options' instead.");
+    }
+
+  ~PonoOptions() {};
+  
+  PonoResult parse_and_set_options (int argc, char ** argv);
+
+  // Pono options
+  Engine engine;
```
ahmed-irfan, line 57: we use the convention to append `_` at the end of every data member.

ahmed-irfan: though, we are not clear if we would use this convention for private members only

ahmed-irfan:
clang-format

utils/ponoresult.h
```diff
@@ -0,0 +1,33 @@
...
+
+#pragma once
+
+#include <string>
+
+namespace pono {
+
+typedef enum
+{
+  PROPERTY_TRUE = 0,
```
makaimann, line 25: It definitely makes sense to avoid magic constants, but I'm wondering if we should just unify this and ProverResult instead of having two separate classes: https://github.com/upscale-project/pono/blob/e25fb8fa112eeb2b85427fda9475d9b247994654/core/proverresult.h#L25

options/options.cpp
```diff
@@ -0,0 +1,199 @@
...
+    "  --vcd \tName of Value Change Dump (VCD) if witness exists." },
+  { NOWITNESS, 0, "", "no-witness", Arg::None, "  --no-witness \tDisable printing of witness." },
+  { 0, 0, 0, 0, 0, 0 }
+};
+/*********************************** end Option Handling setup
+ * ***************************************/
+
+namespace pono {
+
+bool PonoOptions::instance_created = false;
```
makaimann, line 120: This is global, right? Does this mean there's only one PonoOptions object allowed? We don't have to do this now, but it might actually be nice to allow multiple option configurations in memory. In IC3IA for example, options are a struct that's passed to the prover and it accesses the different options from there. Then, you can have programs that run provers with different options.

options/options.h
```diff
@@ -0,0 +1,73 @@
...
+/*************************************** Options class
+ * ************************************************/
+// Meant to be used as a singleton class
+
+namespace pono {
+
+class PonoOptions
+{
+
+  // TODO: for now, all class members (i.e., Pono options) are public
```
makaimann, line 32: We could also just make this a struct. For this kind of object, I'm not opposed to keeping everything public. What are your thoughts @lonsing and @ahmed-irfan?

lonsing:
Thanks for you feedback, @makaimann and @ahmed-irfan!

I addressed your comments in my changes:

- running clang-format
- member variable names have post-fix "_"
- multiple `PonoOptions` objects can exist. Optionally, a `PonoOptions` object can be passed to the constructor of the `Prover` class, to make potentially different options available in a `Prover`
- merge `ProverResult` and `PonoResult` classes, removing the latter. `ProverResult` now has `ERROR` as an additional value.

engines/bmc.h
```diff
@@ -24,6 +24,7 @@ class Bmc : public Prover
 {
  public:
   Bmc(const Property & p, smt::SmtSolver & solver);
+  Bmc(PonoOptions & opt, const Property & p, smt::SmtSolver & solver);
```
ahmed-irfan, line 27: maybe the options object reference be const

engines/prover.h
```diff
@@ -27,6 +28,7 @@ class Prover
 {
  public:
   Prover(const Property & p, smt::SmtSolver & s);
+  Prover(PonoOptions & opt, const Property & p, smt::SmtSolver & s);
```
ahmed-irfan, line 31: const PonoOptions &opt

options/options.h
ahmed-irfan: struct sounds good.

ahmed-irfan: on a second thought, maybe class would be more clean and safe.

pono.cpp
```diff
@@ -133,123 +133,52 @@ const option::Descriptor usage[] = {
...
-  argc -= (argc > 0);
-  argv += (argc > 0);  // skip program name argv[0] if present
-  option::Stats stats(usage, argc, argv);
-  std::vector<option::Option> options(stats.options_max);
-  std::vector<option::Option> buffer(stats.buffer_max);
-  option::Parser parse(usage, argc, argv, &options[0], &buffer[0]);
+  PonoOptions pono_options;
+  ProverResult res = pono_options.parse_and_set_options(argc, argv);
+  if (res == ERROR) return res;
+  assert(res == pono::UNKNOWN);
```
makaimann, line 173: Could you add one comment before this assert just clarifying that it should be unknown if the options are all set and error otherwise?

lonsing:
Thanks, @makaimann and @ahmed-irfan, I included your suggested changes.

Please let me know whether you are fine with merging this.

PR #52, opened 2020-07-10
interp-mc: fix step function so that it could be used multiple times

ahmed-irfan:


makaimann:
Looks like it fails a test in `test_engines`. I think `R` should still be initialized somewhere.

ahmed-irfan:
`R_` is initialized here https://github.com/upscale-project/pono/blob/022e37169a805c7f79f21ed571bb69f14f081ca9/engines/interpolantmc.cpp#L77

makaimann:
Ahh okay. Just pulled the branch and ran the test locally -- it's unsound. It's because `R_` needs to be re-initialized to `init0_` every time the bound is extended.

ahmed-irfan:
i see. I thought we could use the previous over-approximation in the next run. Therefore, closing this PR

makaimann:
Ahh yeah, unfortunately not. The over-approximation is only for the current unrolling.

PR #53, opened 2020-07-10
mc-interp: refactor

ahmed-irfan:


ahmed-irfan:
I went on refactoring the code to simplify it. This would require a deep review.

makaimann:
Screen Shot 2020-07-13 at 11 40 14 AM
Screen Shot 2020-07-13 at 11 41 47 AM


Looks good! Even picked up a few new benchmarks. There was an error, which looks like it might have been in MathSAT? It was `malloc(): unsorted double linked list corrupted` in `adding.3.prop1-func-interl.btor2`. There's also one error (segfault) in the baseline, so I'm not sure if we should worry about it too much. What do you think?

ahmed-irfan:
I think we can merge.

PR #54, opened 2020-07-12
Allow using a fresh solver in each engine

makaimann:
This PR adds features to support checking multiple properties on the same transition system. Instead of using the same solver everywhere, it adds a TermTranslator that moves terms to a fresh solver. This was implemented as a type of specialized copy constructor in `TransitionSystem` and `Property`. For this change to work, some classes that did not previously "own" a solver now need to own it (i.e. use a shared_ptr directly instead of a reference to a shared_ptr). It took me a little while to get this right, especially with `const`-ness. The default solver for a `TransitionSystem` is CVC4 because it does not rewrite on the fly and supports the most theories.

I added a new test that encodes several `btor2` examples, copies them between `TransitionSystems` with different `SmtSolvers` and tests that the resulting `TransitionSystems` are equivalent. Unfortunately, we cannot rely on structural/syntactic equivalence because the `TermTranslator` object has to do some casting to handle sort-aliasing when transferring to/from Boolector. Thus, I actually use a solver to check that the system is equivalent after being copied to a different solver and then copied back to the original. For the largest system (`ridecore.btor`) I only do this check with Boolector because the other solvers take way too long. If you think the tests are still too slow, I can remove that check.

Finally, the binary is essentially unchanged at this time because I left the solver instantiation the same. In the future, we can make a PR where the binary uses a fresh solver as well (so it can solve multiple properties). However, I would like to test this on the cluster first. I'll create a temporary version that uses a fresh solver in this PR and run it on benchmarks on the cluster to look for bugs and compare runtime. This might take a little while to complete so I'll mark this do-not-merge in the meantime, but you can start reviewing whenever you have time. Thanks!


lonsing:
Hi @makaimann , thanks for opening this PR!

Just a quick question: is the last commit `d271e5b1a551968a2fd64cb4de692ee92ca48023` "Merge branch 'master' into engine-fresh-solver" actually essential to the new functionality of this PR? As far as I see, it merges changes to the SMV frontend.

makaimann:
Hi @lonsing, no it's only necessary for keeping it up to date with master. I'll actually have to merge master again to resolve the latest conflicts from your PR. I'll do that ASAP! You can safely ignore those commits -- the changes shouldn't show up in the Files Changed.

lonsing:
That makes sense, thanks! I just noticed it when reviewing on the command line, where these commits showed up.

makaimann:
Screen Shot 2020-07-15 at 3 39 31 PM
Screen Shot 2020-07-15 at 3 40 12 PM
Screen Shot 2020-07-15 at 3 40 55 PM


Screen Shot 2020-07-15 at 3 40 55 PM
Screen Shot 2020-07-15 at 3 43 38 PM
Screen Shot 2020-07-15 at 3 44 05 PM


Here's the first set of results from the cluster where I ran this branch with a slight modification so that the transition system uses a default solver first. Currently, this is not in the PR. I'm also planning to run a debug version. Looks like there are no discrepancies. Unfortunately, we do lose a few benchmarks (about one per category). I looked into this a bit more deeply on the hwmcc19 bmc configuration. It doesn't seem straightforward. As in, it's not like this branch just didn't solve one of the benchmarks. It solved some that master didn't and didn't solve some that master did. Overall it does slightly worse. This isn't too surprising -- if the benchmark was close to the timeout, translating the transition system could make the difference. Furthermore, encoding to CVC4 first could have changed the transition system slightly in which case all bets are off in terms of search path. I think this is pretty good evidence so far that the changes in this PR are working -- at least without obvious bugs.

lonsing:
Thanks for the update on the results! I agree, it looks pretty good. 

It's interesting that the runs with this branch often have a smaller memory footprint. 

What are the other benchmarks apart from hwmcc19 (the set containing 296 instances)?

makaimann:
Yeah that is interesting -- I don't have an explanation for the smaller memory footprint actually... Unless it's changes separate from this branch since the 10 July 2020 baseline. As for the other benchmark set, it's from my TACAS 2020 paper. They're a bunch of FALSE properties over FIFO-based systems.

core/prop.h
```diff
@@ -25,14 +25,22 @@ class Property
...
+
   ~Property();
 
   const smt::Term prop() const { return prop_; }
 
   const TransitionSystem & transition_system() const { return ts_; }
 
  private:
-  const TransitionSystem & ts_;
+  const TransitionSystem ts_;
```
ahmed-irfan, line 43: what is the motivation for this change here?

makaimann: Ahh that's because the property constructor allows copying the transition system to a new solver. So the transition system needs to be owned by it. 

Thus, when you create a new engine, it copies the property, which also copies the property's transition system. So the engine actually has a reference to the system and the real owner is the new property object.

core/unroller.h
```diff
@@ -43,7 +43,7 @@ class Unroller
   smt::UnorderedTermMap & var_cache_at_time(unsigned int k);
 
   const TransitionSystem & ts_;
-  smt::SmtSolver & solver_;
+  const smt::SmtSolver & solver_;
```
ahmed-irfan, line 46: shouldn't this be a smart-pointer without ref?

engines/interpolantmc.h
```diff
@@ -30,11 +30,10 @@ class InterpolantMC : public Prover
 {
  public:
   // IMPORTANT: assume the property was built using the interpolating solver
-  InterpolantMC(const Property & p, smt::SmtSolver & slv, smt::SmtSolver & itp);
-  InterpolantMC(const PonoOptions & opt,
-                const Property & p,
-                smt::SmtSolver & slv,
-                smt::SmtSolver & itp);
+  InterpolantMC(const Property & p, smt::SolverEnum se);
+  InterpolantMC(const Property & p, smt::SmtSolver slv, smt::SmtSolver itp);
```
ahmed-irfan, line 34: const ref to the SmtSolver

engines/interpolantmc.h
```diff
@@ -30,11 +30,10 @@ class InterpolantMC : public Prover
...
+  InterpolantMC(const PonoOptions & opt, const Property & p, smt::SolverEnum se);
+  InterpolantMC(const PonoOptions & opt, const Property & p, smt::SmtSolver slv, smt::SmtSolver itp);
```
ahmed-irfan, line 36: const ref to SmtSolver

engines/bmc.h
```diff
@@ -23,8 +23,10 @@ namespace pono {
 class Bmc : public Prover
 {
  public:
-  Bmc(const Property & p, smt::SmtSolver & solver);
-  Bmc(const PonoOptions & opt, const Property & p, smt::SmtSolver & solver);
+  Bmc(const Property & p, smt::SolverEnum se);
+  Bmc(const Property & p, smt::SmtSolver solver);
+  Bmc(const PonoOptions & opt, const Property & p, smt::SolverEnum se);
+  Bmc(const PonoOptions & opt, const Property & p, smt::SmtSolver solver);
```
ahmed-irfan, line 29: const ref to smtSolver

engines/bmc.h
```diff
@@ -23,8 +23,10 @@ namespace pono {
 class Bmc : public Prover
 {
  public:
-  Bmc(const Property & p, smt::SmtSolver & solver);
-  Bmc(const PonoOptions & opt, const Property & p, smt::SmtSolver & solver);
+  Bmc(const Property & p, smt::SolverEnum se);
+  Bmc(const Property & p, smt::SmtSolver solver);
```
ahmed-irfan, line 27: const ref to smtSolver

engines/bmc_simplepath.h
```diff
@@ -23,10 +23,10 @@ namespace pono {
 class BmcSimplePath : public KInduction
 {
  public:
-  BmcSimplePath(const Property & p, smt::SmtSolver & solver);
-  BmcSimplePath(const PonoOptions & opt,
-                const Property & p,
-                smt::SmtSolver & solver);
+  BmcSimplePath(const Property & p, smt::SolverEnum se);
+  BmcSimplePath(const Property & p, smt::SmtSolver solver);
```
ahmed-irfan, line 27: const ref to smtSolver

engines/bmc_simplepath.h
```diff
@@ -23,10 +23,10 @@ namespace pono {
...
+  BmcSimplePath(const PonoOptions & opt, const Property & p, smt::SolverEnum se);
+  BmcSimplePath(const PonoOptions & opt, const Property & p, smt::SmtSolver solver);
```
ahmed-irfan, line 29: const ref to smtSolver

engines/kinduction.h
```diff
@@ -23,10 +23,10 @@ namespace pono {
 class KInduction : public Prover
 {
  public:
-  KInduction(const Property & p, smt::SmtSolver & solver);
-  KInduction(const PonoOptions & opt,
-             const Property & p,
-             smt::SmtSolver & solver);
+  KInduction(const Property & p, smt::SolverEnum se);
+  KInduction(const Property & p, smt::SmtSolver solver);
```
ahmed-irfan, line 27: const ref to SmtSolver

engines/kinduction.h
```diff
@@ -23,10 +23,10 @@ namespace pono {
...
+  KInduction(const PonoOptions & opt, const Property & p, smt::SolverEnum se);
+  KInduction(const PonoOptions & opt, const Property & p, smt::SmtSolver solver);
```
ahmed-irfan, line 29: const ref to smtSolver

engines/prover.h
```diff
@@ -27,8 +27,11 @@ namespace pono {
 class Prover
 {
  public:
-  Prover(const Property & p, smt::SmtSolver & s);
-  Prover(const PonoOptions & opt, const Property & p, smt::SmtSolver & s);
+  Prover(const Property & p, smt::SolverEnum se);
+  Prover(const Property & p, smt::SmtSolver s);
```
ahmed-irfan, line 31: const ref to smt solver

engines/prover.h
```diff
@@ -27,8 +27,11 @@ namespace pono {
...
+  Prover(const PonoOptions & opt, const Property & p, smt::SolverEnum se);
+  Prover(const PonoOptions & opt, const Property & p, smt::SmtSolver s);
```
ahmed-irfan, line 33: const ref to smtsolver

core/ts.h
```diff
@@ -170,37 +194,175 @@ class TransitionSystem
    * states */
   bool no_next(const smt::Term & term) const;
 
+  // term building functionality -- forwards to the underlying SmtSolver
+  // assumes all Terms/Sorts belong to solver_
```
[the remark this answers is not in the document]
makaimann: We don't check this in smt-switch yet, but that's something I've been wondering if we should check. It might be nice to discuss this further. There's not an easy way to do this at the Pono level either. Terms/Sorts currently don't have a reference to the solver. So we could either add a reference to the solver in smt-switch or have a special function that needs to be implemented which allows you to check.

core/ts.cpp
```diff
@@ -14,13 +14,81 @@
...
 using namespace std;
 
 namespace pono {
 
+TransitionSystem::TransitionSystem(const TransitionSystem & other_ts,
+                                   TermTranslator & tt)
+{
+  function<Term(const Term &)> transfer;
+  function<Term(const Term &, SortKind)> transfer_as;
+  if (other_ts.solver() == tt.get_solver()) {
```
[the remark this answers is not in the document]
makaimann: Yep, that's the intended behavior! Underlying solvers aren't expected to be able to share terms between different instances, even if they're the same type of solver.

engines/prover.h
```diff
@@ -40,10 +43,14 @@ class Prover
   ProverResult prove();
 
  protected:
-  const TransitionSystem & ts_;
-  const Property & property_;
+  smt::SmtSolver solver_;
+  smt::TermTranslator to_prover_solver_;
+  const Property property_;
+  const TransitionSystem &
```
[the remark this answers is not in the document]
makaimann: Yeah definitely, this is just for convenience. We could remove it if you prefer?

core/ts.h
```diff
@@ -127,11 +152,11 @@ class TransitionSystem
   bool is_next_var(const smt::Term & sv) const;
 
   // getters
-  smt::SmtSolver & solver() { return solver_; };
+  const smt::SmtSolver & solver() const { return solver_; };
```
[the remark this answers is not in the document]
makaimann: Yeah that's exactly right.

core/prop.h
```diff
@@ -25,14 +25,22 @@ class Property
...
 
   smt::Term prop_;
```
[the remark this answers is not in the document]
makaimann: Good catch. Done with 082ef27d73c982ea26ec49d5b0dffc497080e0f6.

core/unroller.h
makaimann: Ahh right we said we would not use references for solvers anywhere, right? Changed with 397a95a03d3e580fde5604e398a3c0d21b831dfb

engines/interpolantmc.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

engines/bmc.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

engines/bmc_simplepath.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

engines/kinduction.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

engines/kinduction.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

engines/prover.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

engines/prover.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

engines/bmc_simplepath.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

engines/bmc.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

engines/interpolantmc.h
makaimann: Thanks, addressed with 0dbc624d1d5d9ab57c7977b758211620fdff5ecd

makaimann:
All right, I addressed all the comments by making changes or responding I think! There are a few things that aren't really resolved, but I'm thinking they could be addressed in a separate PR/discussion. In particular, I think it would be good to discuss how to best check that terms/sorts match the solver.

The debug versions of the jobs also finished, and the results are here. Looks good to me.

Screen Shot 2020-07-20 at 7 17 15 AM
Screen Shot 2020-07-20 at 7 19 04 AM

makaimann:
Hey @lonsing and @ahmed-irfan -- I just updated this branch to keep it inline with master. Sorry I forgot -- where were we on this again? Do you think it's ready to merge or should we make other changes?

lonsing:
Hi @makaimann, this looks good to me!

Some points that we wanted to discuss after merging:
- quoting from one of your comments: `In particular, I think it would be good to discuss how to best check that terms/sorts match the solver.`
- it would be good to do some more systematic testing of the TS copying functionality in the style of `tests/test_btor2_ts_copy_equal.cpp`; I have that on my list.

makaimann:
Great, thanks! Just updated it for changes in master. Everything passed so I'll merge it now.

PR #56, opened 2020-07-15
Keep track of constraints added to trans through add_constraints, add_invar, and constrain_inputs

makaimann:
This PR would add an extra vector of terms to `TransitionSystems` that records constraints added to the transition relation. Currently, those constraints are just added directly to the `trans` formula. However, any future passes (such as COI or fault injection) that modify the transition relation and rebuild it might want to rebuild the transition relation. For a functional system, you could now rebuild `trans` by AND-ing together all the equalities from `state_updates_` and the constraints in `constraints_`.

### DIFF
--- a/core/ts.h
+++ b/core/ts.h
@@ -157,6 +157,12 @@ class TransitionSystem
     return named_terms_;
   };
 
+  /** @return the constraints of the system
+   *  Note: these do not include next-state variable updates or initial state
+   * constraints
+   */
+  const smt::TermVec & constraints() const { return constraints_; };
+
   /** Whether the transition system is functional
    *  NOTE: This does *not* actually analyze the transition relation
    *  it only returns true if it's a FunctionalTransitionSystem object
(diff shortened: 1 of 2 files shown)


core/ts.h
```diff
@@ -201,6 +207,9 @@ class TransitionSystem
   // maps next back to curr
   smt::UnorderedTermMap curr_map_;
 
+  smt::TermVec constraints_;  ///< constraints added via
```
lonsing, line 210: Just a suggestion: maybe you could add some of the description of this PR as comment before this line, which will be valuable documentation, like: "extra vector of terms to TransitionSystems that records constraints added to the transition relation." and "For a functional system, you could now rebuild trans by AND-ing together all the equalities from state_updates_ and the constraints in constraints_."

makaimann: Great idea! Made the change with e0286ad16b965de5c7adbed1e99243314cb8a0d8

PR #57, opened 2020-07-18
Basic IC3 Implementation

makaimann:


makaimann:
@ahmed-irfan Here are the latest results. It did better than I expected for not having _any_ generalization.

Screen Shot 2020-07-30 at 11 14 15 AM

engines/mbic3.h
```diff
@@ -24,26 +24,19 @@
...
-  smt::Term term_;     // term representation of literals as disjunction
-};
-
-struct Cube
-{
-  Cube() {}
-  Cube(const smt::SmtSolver & solver, const smt::TermVec & lits);
+  Conjunction() {}
+  Conjunction(const smt::SmtSolver & solver, const smt::TermVec & lits);
   smt::TermVec lits_;  // list of literals sorted by hash
```
ahmed-irfan, line 35: this should be conjunts_

ahmed-irfan: also in the constructor

engines/mbic3.cpp
```diff
@@ -676,7 +676,7 @@ size_t ModelBasedIC3::push_blocking_clause(size_t i, Term c)
 
   Result r;
   size_t j;
-  for (j = i; j < frames_.size(); ++j) {
+  for (j = i; j + 1 < frames_.size(); ++j) {
```
ahmed-irfan, line 679: can you check this @makaimann ?

engines/mbic3.h
makaimann: Good catch! Fixed with eb808a090671e330a379b06844ee32f975a69e87

engines/mbic3.cpp
makaimann: Yep that looks good to me! Should be equivalent to the previous fix but this is much cleaner, thanks

makaimann:
Here are the latest runs. Still looking good. Looks like we're not adding any unsat benchmarks anymore but still picking up more sat ones.
Screen Shot 2020-08-27 at 7 49 11 AM

makaimann:
Here are the latest results. Looks like the one-shot inductive generalization is actually worse in performance (at least on these benchmarks). The jobs are tagged with the first 5 characters of the hash, so the latest one is the `b0c40` one. I'm not sure that means it's necessarily worse for the interpolant generation though. What do you think?
Screen Shot 2020-08-28 at 8 42 47 AM

Issue #64, opened 2020-08-06
Unexpected sort for `neq` operation on arrays


nakengelhardt:
I was testing the following BTOR2 model, which is doing an equality check on two uninitialized arrays:
```
1 sort bitvec 1
2 sort bitvec 2
3 sort bitvec 8
4 sort array 2 3
5 state 4 mem1
6 state 4 mem2
7 neq 1 5 6
8 bad 7
```
And for this model pono returns:
```
Unexpected sort
unknown
b0
```
Instead of the expected witness with an arbitrary element in the array differing.

Curiously enough, if I only change `neq` to `eq` in the model, that works normally:
```
sat
b0
#0
0 [11] 11111111 mem1@0
0 00000000 mem1@0
1 [11] 11111111 mem2@0
1 00000000 mem2@0
@0
.
```

makaimann:
Hi @nakengelhardt, 

Thanks for the bug report! `neq` was incorrectly only in the operator maps for bit vectors and bools. This will be fixed once this [PR](https://github.com/upscale-project/pono/pull/65) is merged.

nakengelhardt:
Thanks for the quick fix!

PR #65, opened 2020-08-06
Handle neq for arrays in btor2 encoder

makaimann:
Addresses https://github.com/upscale-project/pono/issues/64. The `neq` btor2 tag was originally included in the BV and Bool operator maps but did not have an implementation for arrays. This PR just handles this case for arrays as well.


makaimann:
Running some tests on the cluster before merging.

makaimann:
Screen Shot 2020-08-07 at 11 23 11 AM

PR #66, opened 2020-08-10
Always include a transition before intersecting with bad

makaimann:
Instead of intersecting the last frame with bad, always use a transition from the second to last frame.

Initial results look good (posted below), but I haven't yet fully convinced myself this is correct.

### DIFF
--- a/engines/mbic3.cpp
+++ b/engines/mbic3.cpp
@@ -13,8 +13,9 @@
 **        transition system (exploiting this structure for
 **        predecessor computation) and uses models.
 **/
-#include "engines/mbic3.h"
+#include "smt-switch/utils.h"
 
+#include "engines/mbic3.h"
 #include "utils/logger.h"
 
 using namespace smt;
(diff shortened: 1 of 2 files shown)


makaimann:
Existing results. Running some more now after merging the latest changes from the ic3 branch.
Screen Shot 2020-08-10 at 3 48 14 PM

makaimann:
@ahmed-irfan I just added what we discussed about the last frame in `intersects_bad`. The cluster is a bit backed up now, but we can run more tests later before merging if you'd like.

engines/mbic3.cpp
```diff
@@ -207,67 +208,25 @@ ProverResult ModelBasedIC3::step_0()
...
-
-    Conjunction c(solver_, red_lits);
-    add_proof_goal(c, reached_k_ + 1);
+  TermVec conjuncts;
+  conjunctive_partition(bad_, conjuncts);
+
+  // include the last frame
+  // this is an optimization, it is not necessary for correctness
+  assert(frames_.size() == reached_k_ + 2);
+  for (auto c : frames_[reached_k_ + 1]) {
```
ahmed-irfan, line 217: perhaps we could use `frames_.back()`

ahmed-irfan: i don't have a strong preference for what I suggested

makaimann: Ahh good point! Made that change with 9e361e12d41017dda405d5735c49f1d0ff3f354a. I don't have a strong preference either. It seems kind of nice in principle to use it. Maybe we just keep it and if we have time we can evaluate whether it helps or not later?

makaimann:
Here are the latest results -- it's looking pretty good. Adds a few unique solves even relative to other versions of ic3.
Screen Shot 2020-08-13 at 10 35 58 AM

ahmed-irfan:
Looks good. 3 instances might be noise. Is this branch solving the benchmarks faster then the ic3 branch?

makaimann:
Here's the latest results without asserting the last frame. Still missing those 3 unsat benchmarks but it could still be noise like you said. We gain a bunch of sat benchmarks though so I was thinking we should keep it this way.
Screen Shot 2020-08-13 at 6 00 06 PM

PR #68, opened 2020-08-13
Add support for UF to ModelBasedIC3

makaimann:
Adds a walker for getting all terms with a `PrimOp` that is a member of a given set. Use this to find all UF applications in the transition system and include this in the cube representing a state.


makaimann:
Thanks for reviewing!

PR #71, opened 2020-08-14
Add method to get a name for a term and use it for printing property

makaimann:
This PR adds functionality to the `TransitionSystem` to save a mapping from terms to representative names. It's only a representative name, because it's not guaranteed to have a single name. The main use for this is for printing the property. In `BTOR2` each line can be tagged with a name. In that case, it might be more useful to print the name of the property than the term representing it.


core/ts.cpp
```diff
@@ -177,6 +177,8 @@ void TransitionSystem::name_term(const string name, const Term & t)
     throw PonoException("Name " + name + " has already been used.");
   }
   named_terms_[name] = t;
+  // save this name as a representative (might overwrite)
+  term_to_name_[t] = name;
```
lonsing, line 181: Would it make sense to enforce that every term has a unique name? E.g. throw an exception if term `t` already has an entry in `term_to_name_`.

Or is there a particular use case to allow calling `name_term` on the same term multiple times but with different names?

makaimann: Good question! So, `name_term` does throw an exception if you try to use a name more than once. However, as you pointed out, it does allow giving different names to the same term. That's not an unreasonable thing to me -- you might have different "views" of the same signal. However, it does complicate the mapping back from terms to a name. What do you think?

lonsing: Thanks for clarifying! I do not have a strong opinion here. I definitely agree that having different views of the same signal can be useful. So perhaps the best thing to do is to move forward with the current version.

PR #75, opened 2020-09-01
Static cone-of-influence analysis

lonsing:
One-time application of cone-of-influence (COI) analysis. We collect state/input variables that appear 
1. in the term that represents the bad-state property and
2. in the terms that represent all the global constraints that were added to the transition system. 

Then we collect all state/input variables that appear in the next-state functions of all previously collected state variables. This process is iterated until saturation.

Finally, we rebuild the transition relation of the transition system by adding all the global constraints and the next-state functions of the collected state-variables.

COI computation is done in the Prover class in the initialization function. Hence COI can basically be combined with any model checking algorithm (i.e. classes inheriting from Prover). 

A new public function was added to the TransitionSystem class to rebuild the transition relation based on collected COI information (to make that possible, `const` qualifiers were removed e.g. from the transition system member variable in the Property class).

Command line usage: `./pono --static-coi`where verbosity level 1 prints basic information and a statistics summary of remaining state variables, and verbosity level 3 prints in-depth information (recommended for debugging only).

COI currently is only compatible with functional transition system (when called on relational ones, an exception is thrown.)

TO BE ADDED: printing of full witnesses.


engines/prover.cpp
```diff
@@ -68,10 +70,284 @@ void Prover::initialize()
...
+          assert(!ts_.is_next_var(cur));
+          logger.log(3, "collect COI statevar {}", cur);
+          collect_coi_term(new_coi_state_vars, cur);
+        }
+        else if (ts_.inputvars().find(cur) != ts_.inputvars().end()) {
+          assert(!ts_.is_curr_var(cur));
+          assert(!ts_.is_next_var(cur));
+          logger.log(3, "collect COI inputvar {}", cur);
+          collect_coi_term(new_coi_input_vars, cur);
+        }
```
makaimann, line 197: What about the next state variable case? Constraints can have next state variables because if it's an invariant, the constraint is added both over current vars and next vars. Is it safe to ignore those cases because there's always a current state version of the same invariant?

engines/prover.cpp
```diff
@@ -68,10 +70,284 @@ void Prover::initialize()
...
+  Term cur;
+
+  /* Depth-first search of term structure 'term'. */
+  while (!open_terms.empty()) {
+    cur = open_terms.back();
+    open_terms.pop_back ();
+
+    /* Use global set 'coi_visited_terms' here to avoid visiting terms
+       multiple times when we call this function on different terms. */
+    if (coi_visited_terms_.find(cur) == coi_visited_terms_.end()) {
```
makaimann, line 180: If it would be easier, you can use this method to get all the free symbolic (i.e. uninterpreted) constants: https://github.com/makaimann/smt-switch/blob/8725a64ba57ce1cddb822d9a527a5ace223271b4/include/utils.h#L58

core/prop.h
```diff
@@ -24,23 +24,23 @@ namespace pono {
...
    */
-  Property(const Property & prop, smt::TermTranslator & tt);
+  Property(Property & prop, smt::TermTranslator & tt);
 
   ~Property();
 
   const smt::Term prop() const { return prop_; }
 
-  const TransitionSystem & transition_system() const { return ts_; }
+  TransitionSystem & transition_system() { return ts_; }
```
ahmed-irfan, line 40: do we need this?

core/ts.cpp
```diff
@@ -331,6 +331,39 @@ Term TransitionSystem::make_term(const Op op, const TermVec & terms)
...
+  // TODO: check potential optimizations in removing global constraints
+  for (auto constr : constraints_)
+    trans_ = solver_->make_term(And, trans_, constr);
+
+  statevars_.clear();
+  for (auto var : state_vars_in_coi) statevars_.insert(var);
+
+  inputvars_.clear();
+  for (auto var : input_vars_in_coi) inputvars_.insert(var);
+}
```
ahmed-irfan, line 365: are we not going to change state_updates_?

ahmed-irfan: if we don't change state_updates_, we are breaking the assumption that trans() is capturing the state_updates.

engines/prover.cpp
```diff
@@ -68,10 +70,284 @@ void Prover::initialize()
...
+        "COI analysis completed: {} remaining state variables, {} original",
+        statevars_in_coi_.size(),
+        orig_num_statevars_);
+  }
 }
 
 ProverResult Prover::prove() { return check_until(INT_MAX); }
 
+/* For debugging only. */
+void Prover::print_term_dfs(const Term & term)
```
ahmed-irfan, line 102: this could be a static non-member function

engines/prover.cpp
```diff
@@ -68,10 +70,284 @@ void Prover::initialize()
...
+      for (auto child : cur) {
+        cout << "    pushing child: " << child << "\n";
+        open_terms.push_back(child);
+      }
+    }
+  }
+}
+
+/* For debugging only. */
+void Prover::print_coi_info()
```
ahmed-irfan, line 130: this could be a static non-member function

engines/prover.cpp
```diff
@@ -68,10 +70,284 @@ void Prover::initialize()
...
+  for (auto statevar : ts_.statevars())
+    cout << "  " << statevar << "\n";
+
+  cout << "constraints: \n";
+  for (auto constr : ts_.constraints())
+    cout << "  " << constr << "\n";  
+}
+
+/* Add 'term' to 'set' if it does not already appear there. */
+void Prover::collect_coi_term(UnorderedTermSet & set, const Term & term)
```
ahmed-irfan, line 156: this could be a static non-member function

engines/prover.cpp
```diff
@@ -68,10 +70,284 @@ void Prover::initialize()
...
+  /* Seed the search using state-variables that were collected already. */
+  TermVec unprocessed_state_vars;
+  for (auto state_var : statevars_in_coi_)
+    unprocessed_state_vars.push_back(state_var);
+
+  while (!unprocessed_state_vars.empty()) {
+    Term state_var = unprocessed_state_vars.back();
+    unprocessed_state_vars.pop_back();
+    assert(ts_.is_curr_var(state_var));
+    const smt::UnorderedTermMap state_updates = ts_.state_updates();
```
ahmed-irfan, line 222: use a reference

engines/prover.cpp
```diff
@@ -68,10 +70,284 @@ void Prover::initialize()
...
+
+/* Traverse 'term', collect state/input variables, and add them to
+   global sets 'new_coi_state_vars' and 'new_coi_input_vars'. */
+void Prover::compute_term_coi(const Term & term,
+                              UnorderedTermSet & new_coi_state_vars,
+                              UnorderedTermSet & new_coi_input_vars)
+{
+  assert(term != NULL);
+  TermVec open_terms;
+  open_terms.push_back (term);
```
ahmed-irfan, line 170: can a term contain a next state variable?

engines/prover.cpp
```diff
@@ -68,10 +70,284 @@ void Prover::initialize()
...
+      if (statevars_in_coi_.find(sv) == statevars_in_coi_.end()) {
+        statevars_in_coi_.insert(sv);
+        unprocessed_state_vars.push_back(sv);
+      }
+    }
+    for (auto sv : new_coi_input_vars) {
+      collect_coi_term(inputvars_in_coi_, sv);
+    }
+         
+    new_coi_state_vars.clear();
```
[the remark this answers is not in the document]
lonsing: Good catch! This part was left over from a previous version. I will update it.

engines/prover.cpp
lonsing: >  Is it safe to ignore those cases because there's always a current state version of the same invariant?
Yes, that is exactly what I was thinking. That should work since we collect **all** state variables in **all** constraints unconditionally, i.e., without checking whether they are syntactically related to the bad-state term.

engines/prover.cpp
lonsing: Yes, `term` could contain a next-state variable depending on the context in which this function is called. 

- if `term` is the bad-state term, then it cannot contain a next-state variable.
- if `term` is a next-state function, then it cannot contain a next-state variable because the transition system is functional (not yet supporting relational ones).
- if `term` is a constraint, then it could contain a next-state variable, which we ignore because there exists always the corresponding constraint over current-state variables, which we will collect. That is related to the comment by @makaimann above.

engines/prover.cpp
lonsing: Thanks for the pointer to that function, I was not aware of it!

As far as I see, `get_free_symbolic_consts(...)` uses a local cache to avoid visiting terms multiple times, whereas `coi_visited_terms_` is a global cache that is maintained during the entire COI computation. That would make it a bit tricky to achieve the same functionality using `get_free_symbolic_consts(...)`, I think.

core/prop.h
lonsing: We need it for the current version but we will refactor the design and revert that change.

engines/prover.cpp
lonsing: Good catch! I will update it.

engines/prover.cpp
lonsing: I agree. I suggest to leave it for now as it will change during refactoring.

engines/prover.cpp
lonsing: Like above, I agree. I suggest to leave it for now as it will change during refactoring.

core/ts.cpp
lonsing: Good point! I just pushed an update.

lonsing:
Thanks for your feedback @makaimann and @ahmed-irfan!

I just pushed an update to address your comments.

PR #83, opened 2020-09-16
Increase verbosity of printout for registered signals

makaimann:


makaimann:
@zhanghongce -- hope you've been doing well! I was wondering if you could review this simple PR? I had a case where this logger message was being printed a lot (for inputs). I wanted to decrease the verbosity there but wanted to make sure that wasn't an issue. I see a similar check [here](https://github.com/upscale-project/pono/blob/22ae1c026a8f03292cfb98777ef19dfa782ddf9c/printers/vcd_witness_printer.cpp#L293) but that one throws an exception.

zhanghongce:
Hi Makai, thanks for your message! To be short, I think it should be okay.
But the long story might be that the symbol names in BTOR2 (generated by Yosys in this case?) are not well handled by the current code.

And I think this problem is complicated. I have seen that different versions of Yosys generate different formats of names (for example some version of Yosys will put `;` and the corresponding Verilog file and line number after the variable name and that will also be parsed by the BTOR frontend and becomes part of the SMT variable name). If the testcase is shareable per license requirement and etc., maybe I can help improve that. Thanks!

makaimann:
I see, thanks for your quick response! I can definitely share the example. I changed the property so that it's violated quickly.

[example.tar.gz](https://github.com/upscale-project/pono/files/5240302/example.tar.gz)

zhanghongce:
Thanks for sharing! I think my guess might be correct. There are things like this
 `2 input 1 pipe.qed_exec_dup ; ./ridecore-src-buggy/topsim.v:101.13-112.8|./ridecore-src-buggy/pipeline.v:52.22-52.34`

which I think messed up with the name extraction in the code. 
I will base my fixes on the master branch, is that okay?

makaimann:
I see -- makes sense. Basing fixes on master is perfect (we are in the process of merging things but I wouldn't expect any conflicts). Thanks! And no rush btw.

zhanghongce:
Sorry that I think I'm wrong. It looks like things after `;` are indeed not taken in by the Btor2 frontend. I think the issue is related to the fact that more things are added to named_terms now.

#85

makaimann:
Better addressed by https://github.com/upscale-project/pono/pull/85

PR #85, opened 2020-09-17
VCD Printer Update

zhanghongce:
I think what goes into the `named_terms_` might have changed and VCD printer was not catching up with that change.
This is to adapt it for that.

It also solves issues related to Yosys's generated signal name format in Btor2.

And I made some array related checks less verbose.


makaimann:
Ahh I see, my bad for not updating the vcd printer in line with the `named_terms_` changes. Thanks @zhanghongce! I think I'll put this PR in the queue behind the other concurrent ones, but I'll review it and get it merged soon.

PR #88, opened 2020-09-25
Fix for TS copy assignment and a test for it

makaimann:
Before, the following code would result in a `FloatingPointException`:
```
FunctionalTransitionSystem fts;
fts = FunctionalTransitionSystem(solver);
```

That was because the member variables in the TS had the solver listed before `init_` and `trans_`. The default solver is `CVC4`, so `fts` starts with a `CVC4` smt-switch solver and initializes `init_` and `trans_` to true. But then the copy assignment copies the solver first, followed by the two terms. Since there was only one reference to the initial solver, it goes out of scope and when `init_` and `trans_` are overwritten, their destructor is called which results in the exception due to the null pointer to the CVC4 in CVC4's Term objects: https://github.com/CVC4/CVC4/blob/c59345b93b2ecf3552f5205b312c262a1ae5eab8/src/api/cvc4cpp.h#L1109. This would also be a problem for any solver, because the solver should exist as long as terms from it are present.

The fix used in this PR is to implement copy assignment using the [copy-and-swap idiom](https://stackoverflow.com/questions/3279543/what-is-the-copy-and-swap-idiom). This PR also adds a test to make sure no exception is thrown on copy assignment.


makaimann:
Even if/when this is approved, lets wait to merge it until the other PRs are merged.

makaimann:
Oops, looks like this doesn't work -- I forgot about other issues with ordering. Looks like there's simply destructor issues now. Might have to go with the copy assignment implementation approach. I'll close it for now.

makaimann:
All right, I fixed it by using copy-and-swap (see updated description). I'd be interested in your thoughts on this approach. Additionally, how should the copy assignment operator be tested? Should we check that all member variables were copied correctly somehow? But then if we update the members, the test would have to be updated as well to catch the issue.

lonsing (approved):
Looks good to me! I left one comment above, where some member variable was missing in the `swap` function.

lonsing:
> Additionally, how should the copy assignment operator be tested? Should we check that all member variables were copied 
> correctly somehow? But then if we update the members, the test would have to be updated as well to catch the issue.

Yes, I think that checking that all member variables were copied correctly would be the right approach. Maybe by implementing the `==` operator?

> I'd be interested in your thoughts on this approach.

It's a neat and tidy solution. I was reading a bit about copy-and-swap. It would be interesting to see whether performance could be impacted when copying large transition systems.

makaimann:
Thanks for the review and for catching the missing member variable! I fixed that and implemented `operator==` for the test. That is a good point about large transition systems. In general, I'd like to have a discussion soon about how best to use transition systems. 

Currently in the code, we mostly use references to avoid copies, but this can be cumbersome sometimes, especially when new systems are created. It also makes constructors complicated because often there's a lot of initialization happening there. We can include this in our next discussion. We can also use your profiling https://github.com/upscale-project/pono/pull/94 to evaluate this.

lonsing:
Thanks for your update and including `operator==`, looks good to me! Yes, let's keep references vs. copies in mind for the next discussion.

PR #94, opened 2020-10-13
Add optional profiling via gperftools

lonsing:
This PR adds optional CPU profiling via linking against the `gperftools` library:

[https://github.com/gperftools/gperftools](https://github.com/gperftools/gperftools)

The README file of Pono comes with usage and installation information.

For this PR, a limited form of signal handling was added. In order to correctly finalize the generation of profile data, it is necessary to catch e.g. SIGINT or SIGTERM when Pono is interrupted by an external time out or when the user manually aborts Pono.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ if (WITH_MSAT)
 endif()
 
 if (WITH_PROFILING)
+  find_library(GOOGLE_PERF profiler REQUIRED)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DWITH_PROFILING")
 endif()
 
@@ -205,8 +206,8 @@ if (LIBRT)
   target_link_libraries(pono-lib PUBLIC ${LIBRT})
 endif()
 
-if (WITH_PROFILING)
-  target_link_libraries(pono-lib PUBLIC profiler)
+if (GOOGLE_PERF)
+  target_link_libraries(pono-lib PUBLIC ${GOOGLE_PERF})
 endif()
 
 enable_testing()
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ if (WITH_MSAT)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DWITH_MSAT")
 endif()
 
+if (WITH_PROFILING)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DWITH_PROFILING")
```
makaimann, line 30: You could consider also checking if the library exists during configuration which gives a better warning message than it just failing to link later. It would look something like this:
```
find_library(GOOGLE_PERF profiler REQUIRED)
```

and then when you link it below you should be able to use `${GOOGLE_PERF}`.

I don't feel strongly about this though, so it's up to you.

pono.cpp
```diff
@@ -102,6 +107,29 @@ ProverResult check_prop(PonoOptions pono_options,
   return r;
 }
 
+// Note: signal handlers are registered only when profiling is enabled.
```
makaimann, line 110: I don't know as much about signal handlers as I should. Is there a big performance overhead? If not, maybe we should just register them always? Maybe it would give a more graceful exit. Although I guess I'm not sure exactly what we would do in the handler when we're not profiling.

makaimann (approved):
Looks really good to me. Thanks for adding this! It will be really useful. I just left a couple minor comments to consider, but as far as I'm concerned, it's good to go.

CMakeLists.txt
lonsing: Thanks a lot! I just pushed the change you suggested.

PR #96, opened 2020-10-15
Add support for compiling with external coreir

leonardt:
This adds support for compiling with a standard coreir installation into
/usr/local.  In the future we may consider supporting more general
installation locations using CMake find_package style, but for now this
should cover the basic use case. (For example, we could try to find the
verison of CoreIR distributed with the pycoreir wheel, but the wheel
would need to be updated to include the header files, also the static
linking may be something we need to deal with.)

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -16,6 +16,7 @@ Configures the CMAKE build environment.
 --with-msat             build with MathSAT which has a custom non-BSD compliant license.  (default : off)
                         Required for interpolant based model checking
 --with-coreir           build the CoreIR frontend (default: off)
+--with-coreir-extern    build the CoreIR frontend using an externally installed library (e.g. pycoreir) (default: off)
 --debug                 build debug with debug symbols (default: off)
 --python                compile with python bindings (default: off)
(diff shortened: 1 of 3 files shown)


configure.sh
```diff
@@ -16,6 +16,7 @@ Configures the CMAKE build environment.
 --with-msat             build with MathSAT which has a custom non-BSD compliant license.  (default : off)
                         Required for interpolant based model checking
 --with-coreir           build the CoreIR frontend (default: off)
+--with-coreir-extern    build the CoreIR frontend using an externally installed library (e.g. pycoreir) (default: off)
```
makaimann, line 19:
```suggestion
--with-coreir-extern    build the CoreIR frontend using a globally installed library (e.g. pycoreir) (default: off)
```

I don't feel strongly about this, but thought this wording might be clearer if it's still accurate.

makaimann (approved):
Looks good, thanks for the PR! Only left a minor comment.

leonardt:
Actually, even better, I updated the wording to explicitly specify `/usr/local/lib` since that's hardcoded in the CMakeFile

PR #104, opened 2020-10-20
Introduce monitor state when creating a property

makaimann:
Automatically introduces a monitor state value (and logs a warning for verbosity >= 1) if the property has inputs or next states in it. This was previously done only for the BTOR2 frontend, but now happens internally in a `Property` object. This extends the counterexample trace by one step, but allows us to use any state invariant checking technique (not just unrolling based ones).

One disadvantage of this PR is that it modifies the transition system in place. Thus, if we're checking many properties on the same transition system, we could end up with too many monitor state variables. I've opened an issue for that which we can address later.

@leonardt

### DIFF
--- a/core/prop.cpp
+++ b/core/prop.cpp
@@ -93,6 +93,9 @@ void Property::initialize()
     throw PonoException(
         "Cannot use next in property of a functional transition system.");
   }
+
+  // update prop_
+  prop_ = monitor;
 }
 
 }  // namespace pono
--- a/core/prop.cpp
+++ b/core/prop.cpp
@@ -14,10 +14,13 @@
  **
  **/
 
-#include "smt-switch/utils.h"
-
 #include "core/prop.h"
+
+#include "assert.h"
+#include "core/rts.h"
+#include "smt-switch/utils.h"
 #include "utils/exceptions.h"
+#include "utils/logger.h"
 
 using namespace smt;
 
@@ -26,16 +29,6 @@ namespace pono {
 Property::Property(TransitionSystem & ts, const Term & p, std::string name)
   : ts_(ts), prop_(p), name_(name)
 {
-  const UnorderedTermSet & states = ts.statevars();
-  TermVec free_vars;
-  get_free_symbolic_consts(p, free_vars);
-  for (auto s : free_vars) {
-    if (!ts.is_curr_var(s))
-    {
-      throw PonoException("Property should only use state variables");
-    }
-  }
-
   // find a name if it wasn't provided
   // if no name is associated with it in the ts, then it will just
   // be the to_string of the term
@@ -43,6 +36,7 @@ Property::Property(TransitionSystem & ts, const Term & p, std::string name)
   {
     name_ = ts_.get_name(prop_);
   }
+  initialize();
 }
 
 Property::Property(Property & prop, TermTranslator & tt)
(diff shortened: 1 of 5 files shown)


makaimann:
Ran some tests. There are no discrepancies so I think it looks good. Interpolant-based model checking is solving less than this baseline, but I'm pretty sure that's not related to this (there was another change to interpolant based model checking since that baseline run).

Screen Shot 2020-10-23 at 1 46 46 PM
Screen Shot 2020-10-23 at 1 48 10 PM

core/prop.cpp
```diff
@@ -58,4 +52,50 @@ Property::Property(Property & prop, TermTranslator & tt)
...
+
+  Term monitor;
+  size_t id = 0;
+  while (true) {
+    try {
+      monitor = ts_.make_statevar("_monitor_" + std::to_string(id),
+                                  ts_.make_sort(BOOL));
+      break;
+    }
+    catch (SmtException & e) {
```
lonsing, line 75: I am not very familiar with the `smt-switch` code: is it guaranteed that an `SmtException` will **always** be thrown when `ts_.make_statevar(...)` fails, regardless of the cause of failure? (to avoid getting stuck in the infinite loop)

lonsing (approved):
Looks good to me! I just left two comments.

engines/mbic3.cpp
```diff
@@ -142,6 +142,9 @@ void ModelBasedIC3::initialize()
   frames_.push_back({ ts_.init() });
   push_frame();
 
+  assert(ts_.only_curr(bad_));
+  assert(ts_.only_curr(ts_.init()));
+
```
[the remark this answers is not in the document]
makaimann: That's a very good point! I don't feel strongly either way. What do you think -- should we leave it there for documentation or remove it?

core/prop.cpp
makaimann: Good question -- an `SmtException` is the parent class for all exceptions thrown by `smt-switch`, but it is possible that some other type of exception is thrown in the underlying solver and is not caught correctly by smt-switch. In that case, I believe we still wouldn't be stuck in the infinite loop because we'd get an uncaught exception.

makaimann:
Thanks for the review! Responded to the comments directly and I'm happy to make changes based on how that discussion goes.

engines/mbic3.cpp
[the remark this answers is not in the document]
lonsing: Then I suggest to remove them, just for consistency with the other engine classes that are derived from `Prover`, and which do not have these assertions.

core/prop.cpp
lonsing: I see, that makes sense!

core/prop.cpp
```diff
@@ -58,4 +52,50 @@ Property::Property(Property & prop, TermTranslator & tt)
...
+    }
+  }
+  assert(monitor);
+
+  // monitor starts true
+  ts_.constrain_init(monitor);
+
+  if (ts_.no_next(prop_)) {
+    ts_.assign_next(monitor, prop_);
+    ;
```
ahmed-irfan, line 86:
```suggestion
    
```

engines/prover.cpp
```diff
@@ -148,8 +149,7 @@ void Prover::print_coi_info()
     cout << "  " << statevar << "\n";
 
   cout << "constraints: \n";
-  for (auto constr : ts_.constraints())
-    cout << "  " << constr << "\n";  
+  for (auto constr : ts_.constraints()) cout << "  " << constr << "\n";
```
ahmed-irfan, line 152: is this formatted by git-clang-format?

ahmed-irfan (approved):
LGTM

engines/prover.cpp
makaimann: Yeah it must have been because I didn't intend to change that code otherwise.

PR #107, opened 2020-10-23
Ic3 refactor

makaimann:


engines/mbic3.cpp
```diff
@@ -753,24 +781,33 @@ Conjunction ModelBasedIC3::generalize_predecessor(size_t i,
...
 
 void ModelBasedIC3::assert_frame(size_t i) const
 {
-  solver_->assert_formula(get_frame(i));
+  assert(frame_labels_.size() == frames_.size());
+  Term assump;
+  for (size_t j = 0; j < frame_labels_.size(); ++j) {
+    assump = frame_labels_[j];
+    if (j < i) {
+      assump = solver_->make_term(Not, assump);
```
ahmed-irfan, line 806: this is an optimization, not a requirement. Right?

ahmed-irfan (approved):
LGTM. 

Just check the minor comment I made. (perhaps add the clarification there)

Thanks

engines/mbic3.cpp
makaimann: Yep, that's right! I'll add a comment -- good catch.

PR #108, opened 2020-10-23
Add msat ic3ia backend

makaimann:


engines/msat_ic3ia.cpp
```diff
@@ -0,0 +1,191 @@
...
+**/
+
+#include "engines/msat_ic3ia.h"
+
+#include "assert.h"
+#include "smt-switch/msat_factory.h"
+#include "smt-switch/msat_solver.h"
+#include "utils/logger.h"
+
+using namespace ic3ia;
```
ahmed-irfan, line 25: perhaps, we shouldn't import namespace. It will make the code more readable -- we will know that we are calling ic3ia library methods

engines/prover.cpp
```diff
@@ -377,38 +381,57 @@ bool Prover::witness(std::vector<UnorderedTermMap> & out)
...
       SortKind sk = v->get_sort()->get_sort_kind();
-      Term vi = unroller_.at_time(transfer_to_prover_as(v, sk), i);
-      Term r = solver_->get_value(vi);
-      map[v] = transfer_to_orig_ts_as(r, sk);
+      Term pv = transfer_to_prover_as(v, sk);
+      try {
+        map[v] = transfer_to_orig_ts_as(wit_map.at(pv), sk);
+      }
+      catch (std::exception & e) {
+        success = false;
```
ahmed-irfan, line 406: if `success` is false, we don't need to run the next loop.

engines/prover.cpp
```diff
@@ -377,38 +381,57 @@ bool Prover::witness(std::vector<UnorderedTermMap> & out)
...
+        break;
+      }
     }
 
     for (auto elem : orig_ts_.named_terms()) {
```
ahmed-irfan, line 411:
```suggestion
    if (success) {
      for (auto elem : orig_ts_.named_terms()) {
```

engines/prover.cpp
```diff
@@ -377,38 +381,57 @@ bool Prover::witness(std::vector<UnorderedTermMap> & out)
...
+      Term pt = transfer_to_prover_as(elem.second, sk);
+      try {
+        map[elem.second] = transfer_to_orig_ts_as(wit_map.at(pt), sk);
+      }
+      catch (std::exception & e) {
+        success = false;
+        break;
+      }
     }
   }
```
ahmed-irfan, line 422:
```suggestion
    }
 }
```

ahmed-irfan (commented):
Left some comments

ahmed-irfan:
Also, there are conflicts

ahmed-irfan (approved):

The pr lgtm

engines/prover.cpp
makaimann: Good point! Committed your suggested fix.

PR #109, opened 2020-10-23
Add IC3 via Implicit Predicate Abstraction (IC3IA)

makaimann:
This PR would add an implementation of IC3IA based on the [paper](https://arxiv.org/abs/1310.6847) and the open-source [implementation](https://es-static.fbk.eu/people/griggio/ic3ia/index.html) built on mathsat. Having this implementation will allow us to experiment with different research ideas and tweaks to this algorithm. It currently assumes that you use MathSAT because it relies on interpolation for finding new predicates. However, it would be straightforward to use a different main solver and only use MathSAT for interpolants. Additionally, we can add support for other solvers that support interpolation and/or develop alternative techniques for finding useful predicates.

This implementation inherits the base IC3 implementation (`ModelBasedIC3`), and overrides some methods. This is really nice for avoiding code duplication, but was not always straightforward. In particular, it was not straightforward to figure out which methods needed to be overridden and during the process things would often fail silently (see issues below). 

There are several TODOS left in this code and not everything is elegant. Here are a few issues that we can decide to either resolve now, or in a separate PR:

* predicates are not minimized
* there is no support for predecessor generalization
* setting up the state is strange. This is related to the `initialize` pattern used in lots of classes in this codebase. It is often called in the constructor to initialize the class. However, methods called in constructors can be tricky with inheritance -- it will not work as expected for virtual functions. This causes some issues for engines. For example, `ModelBasedIC3` uses a label for the transition relation that should be related to the concrete transition relation. However, in `IC3IA` it should use the abstract transition relation. We can't set those semantics in `initialize` because it is called in the constructor, and so it won't use the correct version. Instead, I moved it to `set_labels` which gets called at the beginning of `check_until` and supports being called more than once. In general, I think we should think more about `initialize` and inheritance.
* it's not always clear which methods need to be overridden. For example, I had a lot of silently failing issues because the predicate state variables (added to refer to a predicate) are only in the abstract system. So if there was a method in `ModelBasedIC3` that called `ts_.next(predvar)` it would not work correctly because `predvar` is not a variable of `ts_`, instead we needed to call `abs_ts_.next(predvar)`. The way around this was to create a `next` method within IC3 that is virtual and automatically uses the right transition system. However, this is somewhat hacky. I'm not sure what the right interface should be. Should `ts_` be replaced by the abstraction? But then what about when we need the concrete system again?
* creating an abstraction is not straightforward. Right now, you need to create a separate transition system, add state variables and rebuild all the behavior. In particular, I made the `add_statevar(cur, next)` method public so that it could be used to move over state variable symbols that were already created. Otherwise we'd get an error when using `make_statevar` again.

I think most of these should be dealt with in a separate PR, but I wanted to put them here for posterity. We can move some of these to Issues that can be addressed later. The latter problems all boil down to "what's the right interface for abstraction refinement techniques, especially ones built on a standard algorithm that doesn't use abstraction."


ahmed-irfan:
do we need still need this: https://github.com/upscale-project/pono/pull/109/files#diff-7a450128401e22cb2e84341add28eb5aeb66823c86e7a10c6045970fe46f02d4L238-L239

modifiers/implicit_predicate_abstractor.h
```diff
@@ -0,0 +1,94 @@
...
+#include "abstractor.h"
+
+namespace pono {
+
+class ImplicitPredicateAbstractor : public Abstractor
+{
+ public:
+  ImplicitPredicateAbstractor(const TransitionSystem & conc_ts,
+                              TransitionSystem & abs_ts)
+      : Abstractor(conc_ts, abs_ts),
```
ahmed-irfan, line 29: why not move the implementation to its cpp file?

utils/term_analysis.cpp
```diff
@@ -50,4 +52,58 @@ UnorderedTermSet get_free_symbols(const Term & term)
...
+                    const Sort & boolsort,
+                    UnorderedTermSet & out,
+                    bool include_symbols)
+{
+  TermVec to_visit({ term });
+  UnorderedTermSet visited;
+
+  // set of boolean operators
+  // boolean terms with these operators are not predicates
+  unordered_set<PrimOp> boolops({ And, Or, Xor, Not, Implies, Iff, Ite });
```
ahmed-irfan, line 66: I am not sure how do we are extracting predicates in the case of Ite?
It should be different from the other cases, right?

tests/test_term_analysis.cpp
```diff
@@ -0,0 +1,73 @@
...
+    boolsort = s->make_sort(BOOL);
+    bvsort = s->make_sort(BV, 8);
+    funsort = s->make_sort(FUNCTION, { bvsort, bvsort, bvsort });
+    relsort = s->make_sort(FUNCTION, { bvsort, bvsort, boolsort });
+  }
+  SmtSolver s;
+  Sort boolsort, bvsort, funsort, relsort;
+};
+
+TEST_P(TermAnalysisUnitTests, GetPredicates)
```
ahmed-irfan, line 27: should have a case for ITE

engines/ic3ia.cpp
```diff
@@ -0,0 +1,369 @@
...
+using namespace smt;
+using namespace std;
+
+namespace pono {
+
+IC3IA::IC3IA(Property & p, SolverEnum se)
+    : super(p, se), abs_ts_(ts_.solver()), ia_(ts_, abs_ts_)
+{
+  initialize();
+}
```
ahmed-irfan, line 38: I am not sure if I understand the issue with `initialize`.
we can call the initialize method of the super class: `super::initializer()`. Isn't this serve the purpose?

engines/ic3ia.cpp
```diff
@@ -0,0 +1,369 @@
...
+Conjunction IC3IA::get_conjunction_from_model()
+{
+  TermVec conjuncts;
+  conjuncts.reserve(pred_statevars_.size());
+  Term val;
+  for (auto p : pred_statevars_) {
+    if ((val = solver_->get_value(p)) == true_) {
+      conjuncts.push_back(p);
+    } else {
+      assert(val == false_);
```
ahmed-irfan, line 195: isn't false depend on the solver? -- we use two different solvers in mbic3.

engines/ic3ia.cpp
```diff
@@ -0,0 +1,369 @@
...
+  if (!init_label_) {
+    // frame 0 label is identical to init label
+    init_label_ = frame_labels_[0];
+  }
+  if (!trans_label_) {
+    trans_label_ = solver_->make_symbol("__trans_label", boolsort);
+    // Note: Using iff instead of implies so that we can negate it
+    // for example in relational preimage generalization
+    solver_->assert_formula(
+        solver_->make_term(Iff, trans_label_, abs_ts_.trans()));
```
ahmed-irfan, line 216: we changed it to `implies` in the master.

ahmed-irfan: shouldn't we also change here?

engines/ic3ia.cpp
```diff
@@ -0,0 +1,369 @@
...
+  // create a new state variable for the IC3 procedure
+  // used to represent the predicate
+  // NOTE: don't need to actually add to the TS
+  //       kept separate for predicate abstraction
+  // TODO: add infrastructure for getting a fresh symbol -- might fail if name
+  // already exists
+  string name = "__pred" + std::to_string(pred_statevars_.size());
+  Term pred_state = abs_ts_.make_statevar(name, boolsort_);
+  pred_statevars_.push_back(pred_state);
+  predvar2pred_[pred_state] = pred;
```
ahmed-irfan, line 269: why do we need to introduce new boolean predicate variables?

ahmed-irfan (commented):
Did a pass and left some comments. 

There are some points that need to be discussed in detail -- perhaps in person.

modifiers/implicit_predicate_abstractor.h
makaimann: Good point -- changed in 54e716aabee3cff707cfdc5a22c601530234616a

engines/ic3ia.cpp
makaimann: Yeah there's a solver and an interpolator, the `false_` and `true_` here are for the main `solver_`, so this should be fine. Although, I see your point. Managing multiple solvers and their terms has been kind of a pain.

engines/ic3ia.cpp
makaimann: Yeah good catch -- I missed it. Updated with 54e716aabee3cff707cfdc5a22c601530234616a.

engines/ic3ia.cpp
makaimann: To be honest, I'm just basing it entirely off the available open-source implementation which introduces indicator variables like that. I don't see any reason why it wouldn't work if we just use the predicates themselves, so I can make that change. It would be simpler.

engines/ic3ia.cpp
makaimann: Right, so I just think it's kinda messy. Basically we have this class hierarchy: `Prover -> MBIC3 -> IC3IA`. You can instantiate `MBIC3` on its own, so it needs to have an `initialize` in its constructor. `IC3IA` also has to have its own `initialize` in its constructor. They're similar, but might initialize things slightly differently. One example difference is that the trans label needs to refer to either the concrete or abstract transition relation. However, in the `MBIC3::initialize` method, you can't know if you're instantiating a "raw" `MBIC3` or if it's just a base for an `IC3IA` class. I.e. `virtual` method calls in a base class constructor _don't_ work as expected. We can discuss this more offline.

makaimann:
Thanks for the feedback! I agree that there are somethings that we should discuss offline.

Regarding this,

> do we need still need this: https://github.com/upscale-project/pono/pull/109/files#diff-7a450128401e22cb2e84341add28eb5aeb66823c86e7a10c6045970fe46f02d4L238-L239

We do still need those `add_statevar` and `add_inputvar` methods to be public because they're used here: https://github.com/upscale-project/pono/blob/54e716aabee3cff707cfdc5a22c601530234616a/modifiers/implicit_predicate_abstractor.cpp#L51. I'm sure there's a cleaner solution, but I'm not sure what it is. The main issue is if we're creating an abstraction, we usually just use the same solver, but then we have to have some way to move state variables over without re-declaring them.

tests/test_term_analysis.cpp
makaimann: Good idea -- added with 2d2e8719a4b54581d7961f391dba6d9d92fd2201. One question about that. If I have something like `x' = ite(x < 5, x+1, 0)`, would you expect the predicates to be `{x < 5 }` or `{x < 5, x' = ite(x < 5, x+1, 0)}`? I.e. does the equality between the term and the ITE-term count as a predicate? Currently it's the latter.

utils/term_analysis.cpp
makaimann: I think it's the same. In this case, `boolops` are the operators that are ignored and so it just recurses on all the children of that term. Although there might be a slight difference depending on what we count as predicates, see comment below.

tests/test_term_analysis.cpp
makaimann: We discussed offline and decided that the semantics of including the equality is correct, but it might be more performant to take the different cases: `x' = x+1` and `x' = 0`. This shouldn't be hard to implement so I will add this to the PR soon.

engines/ic3ia.cpp
makaimann: We decided offline to use a naming schema to make this clearer. Implemented in 790a353ed8f90ad41f186af4e020be7402c6b490.

engines/ic3ia.cpp
makaimann: We decided offline to not use the indicator variables here, made the change with a1501f856fa66adab0e293b53f0ea121a22168a1.

utils/term_analysis.cpp
makaimann: We discussed offline that the ITEs should be handled specially for performance reasons, i.e. for `y = ite(x < 10, x, 0)` we should find predicates `x < 10`, `y = x` and `y = 0` instead of using the whole equality with an ite. Made that change with 612bc97ee8f414511b19af46fdac7436cfe9ce55.

makaimann:
All right, I think I've addressed all the things we discussed offline. I'll start some test runs soon.

makaimann:
Here are the latest results. Some errors that I need to look into. It doesn't do very well on hardware benchmarks, but that's to be expected. Especially because we're not minimizing predicates yet.


Screen Shot 2020-11-11 at 10 43 20 AM

makaimann:
Looks like all the errors were refinement errors.

Screen Shot 2020-11-11 at 10 46 44 AM

makaimann:
@ahmed-irfan noticed I was comparing to a bad baseline, here is a more useful one.

Screen Shot 2020-11-11 at 10 52 00 AM

ahmed-irfan (approved):
LGTM. 

As discussed offline, we can improve it in separate PRs.

makaimann:
Right, improvements will include:
* minimizing predicates
* possibly changing how `initialize` is used
* refactoring the IC3 implementations to be more parameterized

PR #113, opened 2020-11-03
Fix in IC3 relational preimage generalization

makaimann:
Looks like I introduced a bug in https://github.com/upscale-project/pono/pull/107. I didn't catch it until now because it only happens for relational systems. I ran it on hwmcc but I must have not rerun it after making more changes. This PR would fix the bug and modify a test so that this kind of issue is tested for in the future.

It seems that the issue is related to preimage generalization for relational systems where we take the whole formula and negate it. If we use the transition system label instead of the actual transition relation, it causes strange over-generalizations. I don't fully understand it, but I think it's because the label is too weak, we want to include the whole transition relation.

### DIFF
--- a/engines/mbic3.cpp
+++ b/engines/mbic3.cpp
@@ -163,14 +163,13 @@ void ModelBasedIC3::initialize()
 
   // set labels
   Sort boolsort = solver_->make_sort(BOOL);
-  // Note: we're using Iff instead of Implies for the labels so that if we want
-  // to negate the formula we can -- e.g. for generalization purposes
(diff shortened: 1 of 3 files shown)


makaimann:
Actually, another fix is to use `<->` instead of `->` for the trans and init labels, then negating the label will force negating the formula as well. I switched it to that fix because I'd rather use the labels everywhere.

ahmed-irfan (approved):
LGTM.

Thanks for making the changes after our discussion (offline).

PR #115, opened 2020-11-06
Cython pointer cleanup

makaimann:
Because many classes in this codebase do not have nullary constructors, the Cython bindings often use a pointer to the C++ object. These pointers need to be deallocated when the Cython object is destroyed. This was not happening before leading to a memory leak. This PR deallocates the memory by deleting the pointer in the Python method `__dealloc__` of each wrapper class that keeps a pointer.


lonsing (approved):
Looks good to me!

PR #116, opened 2020-11-08
Refactor Cone Of Influence analysis into a class

makaimann:
Also expose it as a function for Python API.


lonsing (commented):
Thanks a lot for working on this refactoring step, @makaimann! Looks good to me.

I did a comparison of the numbers of state/input variables in the COI as computed by this branch and the master branch. For the HWMCC'20 benchmarks, these numbers always match.

However, the refactored version tends to be a bit slower than the previous one on master. For most instances of HWMCC'20 this seems to be negligible (e.g. a fraction of a second vs one second). To measure the time, I ran `./pono -v 1 --static-coi -e bmc -k 0 <instance>` to focus the run on COI computation.

The `ridecore.btor` instance is an outlier -- here, the refactored version runs approx 33s and the original one less than a second. 

To look further into that, I did a profiling run. The refactored version spends 96% of the run time in function `pono::TransitionSystem::curr`, which calls `smt::BoolectorSolver::substitute`.

We call `pono::TransitionSystem::curr` when we rebuild the `named_terms_` based on the COI information. 

Maybe we could use caching inside `pono::TransitionSystem::curr` to resolve that performance bottleneck? In general, are there other places where we call that function frequently?

If it is urgent, we can merge this branch and get back to this issue later. My hope is that this performance bottleneck occurs on very large transition systems only; `ridecore.btor` has by far the largest number of state variables, i.e., >6000, in the HWMCC'20 set.

core/ts.cpp
```diff
@@ -445,6 +447,30 @@ void TransitionSystem::rebuild_trans_based_on_coi(
...
+  TermVec free_vars;
+  for (auto elem : named_terms_) {
+    free_vars.clear();
+    get_free_symbolic_consts(elem.second, free_vars);
+    bool any_in_coi = false;
+    Term currvar;
+    for (auto v : free_vars) {
+      // look at current version of variables (if it contains next states)
+      // NOTE: it could also be an input variable
+      currvar = curr(v);
```
lonsing, line 462: I continued to look further into the performance bottleneck on the `ridecore.btor` instance from HWMCC'20.

Is it actually necessary to call `curr(v)` here, which calls the substitute-function of the SMT solver? It seems to me that `v` should be either an input-, a current-state-, or a next-state-variable. Shouldn't then all the information to obtain the current version of state variables already be available in the `next_map` and `curr_map`?

core/ts.cpp
```diff
@@ -445,6 +447,30 @@ void TransitionSystem::rebuild_trans_based_on_coi(
...
+          || input_vars_in_coi.find(currvar) != input_vars_in_coi.end()) {
+        any_in_coi = true;
+        break;
+      }
+    }
+    if (any_in_coi) {
+      reduced_named_terms[elem.first] = elem.second;
+    }
+  }
+  named_terms_ = reduced_named_terms;
```
lonsing, line 473: Should we also rebuild the `term_to_name_` map here? (I've overlooked that in my review.)

makaimann: Oops good catch, thanks! Fixed that with 287edc930e5e970878594934fa246af22abae505 and c46819fbed0c9d41515562425f959922b936fffb.

core/ts.cpp
makaimann: Yeah that's a great observation, I changed it to just use the `curr_map_` cache with this 4c73f8cda94981fe91d88b15fde7f28eedaec547. I also made an [issue on smt-switch](https://github.com/makaimann/smt-switch/issues/126) to look more into this cause the massive difference in performance is surprising to me.

makaimann:
Thanks for the very thorough review -- I really appreciate it! Especially nice catch on the performance issues. I addressed the `curr` substitute issue by just checking the cache as you suggested and it now runs quickly on `ridecore.btor` on my machine.

lonsing (approved):
Thanks for the updates, @makaimann!

makaimann:
FYI @ntsis this is merged.

PR #119, opened 2020-11-13
Several fixes in IC3IA

makaimann:
This PR fixes a bug in the IC3IA backend and refactors so that we can run counterexample-guided prophecy:
* term transfers from the interpolant solver were failing. The fixes include a change to the A, and B construction as well as copying over UFs to the term translator cache
* as part of that change, it makes use of an updated interface for getting free symbols from smt-switch.
* it factors out the check for supported logics in the TS so that IC3IA can run on systems with uninterpreted sorts, UFs and arrays.


makaimann:
This should not be merged until https://github.com/makaimann/smt-switch/pull/127 is merged. Additionally, the `setup-smt-switch.sh` script should be updated to point at master instead of the branch after that has been merged.

makaimann:
If you'd like to test it, you can run on the simple example with: 
`./pono -v 1 -e ic3ia --ceg-prophecy-arrays ../samples/array_lt200.btor2`

engines/interpolantmc.cpp
```diff
@@ -17,6 +17,7 @@
 #include "engines/interpolantmc.h"
 
 #include "smt-switch/exceptions.h"
+#include "smt-switch/utils.h"
```
ahmed-irfan, line 20: maybe it not needed...

makaimann:
https://github.com/makaimann/smt-switch/pull/127 has been merged and I updated the `smt-switch` setup script to pull the master branch again.

ahmed-irfan (approved):
LGTM overall -- just a minor comment.

engines/interpolantmc.cpp
makaimann: Good thought, I think it is needed in this case for https://github.com/upscale-project/pono/blob/cdb9df7350fe29a52876191c97c4feed471e7dca/engines/interpolantmc.cpp#L97

makaimann:
Thanks for the review!

PR #122, opened 2020-11-20
Refactor IC3 implementations to common, parameterized structure

makaimann:


makaimann:
Hey @ahmed-irfan, I thought I'd make a draft PR where we can discuss the IC3 updates. What do you think of this base class? I started with templates but it was a bit of a pain and I thought maybe we didn't really need it.

Also, I remembered why we had frames over terms instead of the main unit. It's because Frame 0 (e.g. initial states) might not be in that form. We could shift all the frames and never store init in the structure. Or we could just use frames over terms. What do you think?

ahmed-irfan:
> Hey @ahmed-irfan, I thought I'd make a draft PR where we can discuss the IC3 updates. What do you think of this base class? I started with templates but it was a bit of a pain and I thought maybe we didn't really need it.
> 
> Also, I remembered why we had frames over terms instead of the main unit. It's because Frame 0 (e.g. initial states) might not be in that form. We could shift all the frames and never store init in the structure. Or we could just use frames over terms. What do you think?

I think it looks good

engines/ic3base.h
```diff
@@ -0,0 +1,428 @@
...
+struct IC3Goal
+{
+  // based on open-source ic3ia ProofObligation
+  IC3Formula target;
+  size_t idx;
+  // TODO: see if we can make this a unique_ptr
+  //       made it complicated to move from this struct to another place
+  std::shared_ptr<IC3Goal> next;
+
+  IC3Goal(IC3Formula u, size_t i, std::shared_ptr<IC3Goal> n)
```
zhanghongce, line 83: Why not `const std::shared_ptr<IC3Goal> &`? So you can save one copy in the argument.

makaimann: Great catch! That must have been overlooked because it was a unique pointer at one point that got moved instead of copied. Addressed your comment with d3bd2b4959000b9782a5e431d007b4ae0d549a39

makaimann:
All right, I finished moving all the IC3 implementations to this new infrastructure so I think we're ready for reviews. There are still TODOs in the code, some of which we can address now and others can probably be a separate PR later. I think it would be good to evaluate this against the `hwmcc2020` branch and maybe `master` as well since it's a big change, so I've marked this `do not merge` for now. I'll start evaluation runs tomorrow.

engines/ic3.cpp
```diff
@@ -0,0 +1,299 @@
...
+          lits = new_tmp;
+          break;  // next iteration
+        }
+      }
+    }
+
+    progress = lits.size() < prev_size;
+  }
+
+  // TODO: would it be more intuitive to start with a clause
```
makaimann, line 213: @ahmed-irfan flagging this TODO for you. Do you think it would be a cleaner interface for `inductive_generalization` to take a disjunction instead of a conjunction? E.g. take something that's already aligned with the frame and generalize it.

engines/ic3.cpp
```diff
@@ -0,0 +1,299 @@
...
+  //       and generalize the clause directly?
+  IC3Formula blocking_clause =
+      ic3_formula_negate(ic3_formula_conjunction(lits));
+  assert(blocking_clause.is_disjunction());  // expecting a clause
+  return { blocking_clause };
+}
+
+IC3Formula IC3::generalize_predecessor(size_t i, const IC3Formula & c)
+{
+  // TODO: change this so we don't have to depend on the solver context to be
```
makaimann, line 223: My point here is it would be nice if we didn't have these assumptions about the solver state. I didn't want someone implementing the derived class to have to pop the solver context, so that's done after calling this method instead of in it. But that also means that you can't use the solver in `generalize_predecessor` because it's still at context level 1 and has assertions from the relative inductiveness check.

engines/ic3base.h
```diff
@@ -0,0 +1,465 @@
...
+        disjunction(other.disjunction)
+  {
+  }
+
+  virtual ~IC3Formula() {}
+
+  /** Returns true iff this IC3Formula has not been initialized */
+  bool is_null() const { return (term == nullptr); };
+
+  bool is_disjunction() const { return disjunction; };
```
makaimann, line 75: To avoid over-generalizing, decided to just assume that an `IC3Formula` is always a disjunction or conjunction. The advantage of this is that it allows more work to be done in the base class instead of relying on the derived classes to do it. For example, this allows for using the unsat core for a first round of generalization if you know the structure. This could be changed later.

engines/prover.h
```diff
@@ -25,6 +25,17 @@
 #include "smt-switch/smt.h"
 
 namespace pono {
+
+/** enum for communicating result of a refinement step
+ *  only used for algorithms that use abstraction refinement
+ */
+enum RefineResult
```
makaimann, line 32: These are just supposed to be used by CEGAR loops to communicate how refinement went. In IC3Base, `refine` will return `REFINE_NONE` by default -- meaning no refinement is necessary, so a counterexample is always real unless this is overridden.

options/options.cpp
```diff
@@ -48,8 +48,8 @@ enum optionIndex
   NO_IC3_PREGEN,
   NO_IC3_INDGEN,
   IC3_GEN_MAX_ITER,
-  IC3_INDGEN_MODE,
   IC3_FUNCTIONAL_PREIMAGE,
+  MBIC3_INDGEN_MODE,
```
makaimann, line 52: Changed this because that option is only used in `mbic3`.

makaimann (commented):
Left a few comments to point out and explain some things.

engines/ic3.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+  UnorderedTermSet keep, core_set;
+  TermVec bool_assump, tmp, new_tmp, removed, lits;
+  lits = c.children;
+
+  if (options_.random_seed_ > 0) {
+    shuffle(
+        lits.begin(), lits.end(), default_random_engine(options_.random_seed_));
+  }
+
```
ahmed-irfan, line 119:
```suggestion
//TODO: use unsatcore-reducer
```

engines/ic3.cpp
ahmed-irfan: i see most algorithms written in papers use cube as input. So, i would stick to that notation. 
Also, imagine, if we don't do any generalization, then this function would return the negation of the cube.

engines/ic3ia.h
```diff
@@ -19,57 +19,90 @@
...
-  IC3IA(const PonoOptions & opt, Property & p, const smt::SmtSolver & slv);
-  virtual ~IC3IA();
-
-  typedef ModelBasedIC3 super;
+  // itp_se is the SolverEnum for the interpolator
+
+  IC3IA(Property & p,
+        smt::SolverEnum se,
+        smt::SolverEnum itp_se = smt::SolverEnum::MSAT);
+  IC3IA(Property & p,
```
ahmed-irfan, line 46: add empty line space

engines/ic3ia.h
```diff
@@ -19,57 +19,90 @@
...
+        const smt::SmtSolver & s,
+        smt::SolverEnum itp_se = smt::SolverEnum::MSAT);
+  IC3IA(const PonoOptions & opt,
```
ahmed-irfan, line 49: add  empty line space

engines/ic3ia.h
```diff
@@ -19,57 +19,90 @@
...
+        Property & p,
+        smt::SolverEnum se,
+        smt::SolverEnum itp_se = smt::SolverEnum::MSAT);
+  IC3IA(const PonoOptions & opt,
```
ahmed-irfan, line 53: empty line space

engines/ic3ia.h
```diff
@@ -19,57 +19,90 @@
...
+        Property & p,
+        const smt::SmtSolver & s,
+        smt::SolverEnum itp_se = smt::SolverEnum::MSAT);
+  virtual ~IC3IA() {}
```
ahmed-irfan, line 57: empty line space

engines/ic3ia.cpp
```diff
@@ -19,278 +19,256 @@
...
     }
+    cache[to_interpolator_.transfer_term(s)] = s;
   }
-}
 
-void IC3IA::check_ts() const
-{
-  // this is a complete No-Op for IC3IA
-  // no current limitations (among major theories)
+  // TODO fix generalize_predecessor for ic3ia
```
ahmed-irfan, line 213: it shouldn't be a problem, I think.  or Maybe I am missing something here.

engines/ceg_prophecy_arrays.cpp
```diff
@@ -93,6 +93,37 @@ CegProphecyArrays::CegProphecyArrays(const PonoOptions & opt,
...
+      reached_k_++;
+    } while (num_added_axioms_);
+
+    Property latest_prop(abs_ts_, solver_->make_term(Not, bad_));
+    PonoOptions opts = options_;
+    // disable static coi because it can't be called more than once on the same
+    // system
+    // TODO: handle this in a better way
+    opts.static_coi_ = false;
+    shared_ptr<Prover> prover =
```
ahmed-irfan, line 118:
```suggestion
//TODO : think about making it use the same prover -- incrementally
    shared_ptr<Prover> prover =
```

ahmed-irfan (changes requested):
Ok. Finally, I managed to do the review. Thanks for being patient :)

The code looks really cool now. 

I left some comments -- don't need to make all the changes, but we should keep the notes / todos

engines/ic3base.h
```diff
@@ -0,0 +1,463 @@
...
+   *                  variable model
+   *  @param nexts - pointer to a vector. If non-null populate with next
+   *                 state variable model
+   *  @return an IC3Formula over current state variables with is_disjunction
+   *          false depending on the flavor of IC3, this might be a boolean
+   * cube, a theory cube, a cube of predicates, etc... AND if inputs non-null,
+   * then include model for inputs, e.g. as equalities if nexts non-null, then
+   * include model for next state vars, e.g. add equalities to vector
+   */
+  virtual IC3Formula get_ic3_formula(smt::TermVec * inputs = nullptr,
```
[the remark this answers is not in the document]
makaimann: What do you think of just `get_model_ic3_formula`? I only say that because they're not necessarily equalities.

engines/ic3base.h
```diff
@@ -0,0 +1,463 @@
...
+  std::vector<std::vector<IC3Formula>> frames_;
+
+  ///< stack of outstanding proof goals
+  std::vector<ProofGoal> proof_goals_;
+
+  // labels for activating assertions
+  smt::Term init_label_;       ///< label to activate init
+  smt::Term trans_label_;      ///< label to activate trans
+  smt::TermVec frame_labels_;  ///< labels to activate frames
+  smt::UnorderedTermMap labels_;  //< labels for unsat cores
```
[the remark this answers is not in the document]
makaimann: Good point! I'll look into that.

engines/ic3base.h
```diff
@@ -0,0 +1,463 @@
...
+        children(other.children),
+        disjunction(other.disjunction)
+  {
+  }
+
+  virtual ~IC3Formula() {}
+
+  /** Returns true iff this IC3Formula has not been initialized */
+  bool is_null() const { return (term == nullptr); };
+
```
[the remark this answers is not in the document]
makaimann: Ahh yeah that would be nice, but then we'd need a solver in `IC3Formula`. Do you think it's worth it?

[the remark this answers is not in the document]
ahmed-irfan: you have a good point. Let's not do it then.

engines/ic3ia.h
makaimann: done in 776b789b56bedbe54bb965348ada5bd2621dd658

engines/ic3base.h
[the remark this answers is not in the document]
ahmed-irfan: sounds good

engines/ic3ia.h
makaimann: done in 776b789b56bedbe54bb965348ada5bd2621dd658

engines/ic3ia.h
makaimann: done in 776b789b56bedbe54bb965348ada5bd2621dd658

engines/ic3ia.cpp
makaimann: So it was failing, e.g. the formula wasn't unsatisfiable. I think it has something to do with the predicate abstraction? I need to look into it more still. Might be why the MathSAT IC3IA has a bail-out option from the generalization. Even the version for non-deterministic systems isn't guaranteed to be unsat I don't think because we're not using model values.

engines/ic3base.h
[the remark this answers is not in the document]
makaimann: Done with a96e40470732592a903a69f6b4a8050abd28ad9d

engines/ic3base.cpp
```diff
@@ -0,0 +1,731 @@
...
+      // optimization: disable the unused constraints
+      // by asserting the negated label
+      assump = solver_->make_term(Not, assump);
+    }
+    assert(assump);  // assert that it's non-null
+    solver_->assert_formula(assump);
+  }
+}
+
+Term IC3Base::get_frame(size_t i) const
```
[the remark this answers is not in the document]
makaimann: Done with 8d6d931f26c785b86ac11b81bebdf59efe3ebcd3

engines/ic3base.h
```diff
@@ -0,0 +1,463 @@
...
+                      std::shared_ptr<ProofGoal> n);
+
+  /** Check if there are common assignments
+   *  between A and B
+   *  i.e. if A /\ B is SAT
+   *  @param A the first term
+   *  @param B the second term
+   *  @return true iff there is an intersection
+   */
+  bool intersects(const smt::Term & A, const smt::Term & B);
```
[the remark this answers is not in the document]
makaimann: Fixed with https://github.com/upscale-project/pono/pull/122/commits/efb4217f60e9fd949c18b32c8db88a2ae636e3e8

engines/ic3base.h
```diff
@@ -0,0 +1,463 @@
...
+
+  /** Check if the term intersects with the initial states
+   *  syntactic sugar for intersects(ts_.init(), t);
+   *  @param t the term to check
+   *  @return true iff t intersects with the initial states
+   */
+  bool intersects_initial(const smt::Term & t);
```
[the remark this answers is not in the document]
makaimann: Fixed with https://github.com/upscale-project/pono/pull/122/commits/efb4217f60e9fd949c18b32c8db88a2ae636e3e8

engines/ic3base.cpp
```diff
@@ -0,0 +1,728 @@
...
+
+// helper functions
+
+/** Less than comparison of the hash of two terms
+ *  for use in sorting
+ *  @param t0 the first term
+ *  @param t1 the second term
+ *  @return true iff t0's hash is less than t1's hash
+ */
+bool term_lt(const smt::Term & t0, const smt::Term & t1)
```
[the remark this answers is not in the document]
makaimann: Done with https://github.com/upscale-project/pono/pull/122/commits/141f0bc5704f038b5b1f495ac7fc6e6fdc6d24aa

engines/ic3base.cpp
```diff
@@ -0,0 +1,731 @@
...
+      }
+    }
+
+    // Most IC3 implementations will have only a single element in the vector
+    // e.g. a single clause. But this is not guaranteed for all
+    // for example, interpolant-based generalization for bit-vectors is not
+    // always a single clause
+    size_t min_idx = frames_.size();
+    for (auto bu : collateral) {
+      // TODO: fix name -- might not be a clause anymore
```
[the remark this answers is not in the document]
makaimann: Good catch -- removed with https://github.com/upscale-project/pono/pull/122/commits/b70823491c951eb76c679501f3bdac9ff33e6470

engines/ic3base.cpp
```diff
@@ -0,0 +1,731 @@
...
+      solver_->make_term(And, conj, u.term);
+      assert(ic3_formula_check_valid(u));
+      assert(ts_->only_curr(u.term));
+    }
+    assert(!intersects_initial(solver_->make_term(Not, conj)));
+  }
+  assert(solver_context_ == 0);
+
+  if (r.is_sat()) {
+    assert(
```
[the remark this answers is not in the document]
makaimann: Done with https://github.com/upscale-project/pono/pull/122/commits/2f153be595b04c432876af62445ae23fd0f60bc9

engines/ic3base.cpp
```diff
@@ -0,0 +1,728 @@
...
+  assert(!init_label_);
+  assert(!trans_label_);
+  // frame 0 label is identical to init label
+  init_label_ = frame_labels_[0];
+  trans_label_ = solver_->make_symbol("__trans_label", boolsort);
+  solver_->assert_formula(
+      solver_->make_term(Implies, trans_label_, ts_->trans()));
+
+  // check whether this flavor of IC3 can be applied to this transition system
+  check_ts();
```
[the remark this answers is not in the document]
makaimann: Good idea! Done with https://github.com/upscale-project/pono/pull/122/commits/bf4ba8a1d350a51f2da4e85a1a036c3b5b971fdf

engines/ic3base.h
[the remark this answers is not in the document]
makaimann: So if we use the `reducer_` for `inductive_generalization` in `IC3` and `ModelBasedIC3`, would we need to get rid of this https://github.com/makaimann/smt-switch/blob/6e6552140d667e02d3ed2ae7b08b3902af051f2b/src/utils.cpp#L206? Cause for inductive generalization, I think you want to drop a literal, then reduce with an unsat core and repeat. But when you drop a literal, it's not guaranteed to be unsat, right?

makaimann:
@ahmed-irfan here's the initial comparison. There are no discrepancies at least but there are still some fairly major issues:

1. There are a bunch of errors in mbic3 which I will look into
2. Our internal ic3ia is much worse than the external one. Maybe we can look into this more together? Do you have any intuition on why that might be? I can think of three possibilities.
  * there's an issue in our bit-level IC3 algorithm
  * we didn't set the right options in mathsat (because we didn't set any besides unsat-cores and models)
  * overhead from smt-switch

param-ic3-comp

makaimann:
I looked into the errors in `mbic3` and that might be related to the poor performance in `ic3ia` as well. That bug should be fixed by dc27b623a08816d1a98e9cf4c707ed2f110fac33. I'll start new jobs soon.

makaimann:
We want to merge this once we've confirmed that the bug is fixed (cluster currently down). And then we can make improvements in a separate PR. We need to focus on improving the ic3-flavor algorithms' performance. We discussed offline some possible causes for slowdowns:
* bug that was just fixed in `propagate` / `constrain_frame`
* using unsat core reducer with default 0 --> e.g. minimal core. in generalize_predecessor could use 2
* options for the `msat_env`
  * could try getting `msat_env` from `msat-ic3ia` so that the configuration is the same
  * could even use a shared env to avoid translation
* doing interpolation outside instead of using sequence interpolant (our outside loop could be more expensive because it needs to re-solve each time)
  * and do we even need sequence interpolants at all

engines/ic3ia.cpp
```diff
@@ -327,13 +327,6 @@ RefineResult IC3IA::refine()
...
-    TermVec red_preds;
-    if (ia_.reduce_predicates(cex, preds_vec, red_preds)) {
-      // reduction successful
-      preds.clear();
-      preds.insert(red_preds.begin(), red_preds.end());
-    }
-
     // add all the new predicates
     bool found_new_preds = false;
     for (auto p : preds) {
```
ahmed-irfan, line 332: this is going to add the predicates before doing the reduction. We don't want that.

ahmed-irfan: fixed in 31b219f0ea644c93f268b02052895575b6fe765c

makaimann: Oops, thanks for catching that.

makaimann:
Found another issue, just fyi @ahmed-irfan. So `ic3ia` doesn't work correctly with `--static-coi`. This is because `ConeOfInfluence` reduction is done in the `Prover::initialize`. However, we compute the abstraction before calling initialize so that we can pass the abstract system as the main `TransitionSystem`. In theory this would just work, but because all the constraints for implicit predicate abstraction, none of the variables can be removed. This should definitely be fixed in the future, but I was thinking that's probably out-of-scope for this PR which is already getting large. So for now I just throw an exception.

makaimann:
Here are the latest results. I think errors are mostly taken care of except for a few here and there for engines using interpolation. We're still not doing as well as `msat-ic3ia` with the native `ic3ia` implementation, although it gets better if we use `msat` or `boolector` with logging (shown in the `*-manual-logging-btor*` runs). I'm thinking we're about ready to merge and we can make incremental improvements (such as using actual sequence interpolants instead of an outer loop) in separate PRs. At this point we're comparable to master so performance can be a separate issue.

What do you think @ahmed-irfan ?

latest-param-ic3

ahmed-irfan:
I think we can merge it and do the improvements in separate PRs.

I point to note is the number of memouts when we are using the logging solver from  smt-switch.

ahmed-irfan (approved):
Let's merge this and work of the improvements in separate PRs.

makaimann:
Sounds good! I agree, this is getting large and it will be easier to make small improvements asynchronously without all the work being on one branch.

makaimann:
For future reference, here were the `ic3ia` runs from the master branch before merging this PR. I.e. this refactor actually fixed some issues in `ic3ia` and improved performance, so we're headed in the right direction. 

master-ic3ia

PR #125, opened 2020-12-04
Refactor so Prover has a pointer to the transition system

makaimann:
Changes the base Prover class so that it keeps a pointer instead of a reference to the transition system. This is so that engines are free to change what the transition system points to. For example, in a CEGAR technique that utilizes some code from a concrete algorithm (e.g. an IC3 implementation that uses an abstract domain), it's most convenient to have the main `ts_` object be the abstraction, because that's what the algorithm operates on. However, abstraction is not available when the prover class is being constructed.

Note: might be worth considering whether notions of abstraction and refinement should be baked into `Prover` and is just a `no-op` by default. It's possible this would be a better `is-a` relationship with class inheritance. Right now it's a bit funny that some algorithms are CEGAR loops but are derived from a class that doesn't know anything about abstraction or refinement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,12 +44,14 @@ script:
   - ./scripts/travis-setup-osx-python.sh || true
   - sudo python3 -m pip install Cython==0.29.21 --install-option="--no-cython-compile"
   - sudo python3 -m pip install pytest
-  - ./travis-scripts/setup-msat.sh --auto-yes
+  # TEMP disable mathsat because website is down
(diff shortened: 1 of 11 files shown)


makaimann:
NOTE: I've noticed that Travis has been failing for the past few days because the mathsat website appears to be down so it can't download the libraries. If it's not up again soon, we can just disable mathsat in CI for now.

ahmed-irfan (approved):
LGTM

PR #126, opened 2020-12-05
change initialize pattern in engines

ahmed-irfan:


makaimann (changes requested):
LGTM, thanks for this change! I like that a user still doesn't need to worry about calling `initialize`, because it's called in `check_until`. It still might be good to have some assertions elsewhere to make sure it's been initialized (to protect against developers adding new engines but forgetting to initialize). But we can do that later. Other things would definitely fail in that case anyway.

My only minor comment is that the CEG Prophecy implementation should also be updated: https://github.com/ahmed-irfan/pono/blob/745558dad9e65dc6e3b0eb8e7f12c5742dab940f/engines/ceg_prophecy_arrays.h#L53

and also the MsatIC3IA implementation (which I implemented in a funky way, sorry):
https://github.com/ahmed-irfan/pono/blob/745558dad9e65dc6e3b0eb8e7f12c5742dab940f/engines/msat_ic3ia.cpp#L34

ahmed-irfan:
Thanks for the review, Makai!
I think I have addressed your comments. So, please check the PR when you have time.

makaimann:
Thanks for the changes! Looks good to me. One thing to mention is that `prove` was implemented in CEG-Prophecy because there was a slight difference between calling `prove` of the underlying model checker and `check_until`. But I think it's cleaner to not have it there anyway, so let's leave it for now. We can also put it back in later if it becomes important. The better long term solution is to not destroy the underlying prover in each iteration.

makaimann (approved):
LGTM, thanks for the PR and changes!

PR #127, opened 2020-12-11
Add cvc4 option to --smt-solver

makaimann:


ahmed-irfan (approved):
LGTM.

One question: is the transition system going to have one solver, when the cvc4 option is used?

makaimann:
Yeah that's right. In fact, running from the command line it always uses only one solver.

PR #132, opened 2020-12-16
get free variables from formula and unwrap cvc4 terms

makaimann:


engines/ic3ia.cpp
```diff
@@ -485,10 +486,21 @@ bool IC3IA::cvc4_find_preds(const TermVec & cex, UnorderedTermSet & out_preds)
...
+
+  // get the underlying CVC4 objects
+  cvc4a::Term cvc4_formula =
+      static_pointer_cast<CVC4Term>(ss_cvc4_formula)->get_cvc4_term();
+
+  vector<cvc4a::Term> cvc4_free_vars;
+  for (auto fv : ss_free_vars) {
+    cvc4_free_vars.push_back(
+        static_pointer_cast<CVC4Term>(fv)->get_cvc4_term());
+  }
```
yoni206, line 503: I'm not sure that's enough -- don't we need also all the new variables? everything in 
https://github.com/yoni206/pono/blob/13955a8052c3c211cfa7921eca70b90f897f4350/engines/ic3ia.cpp#L591

?

makaimann: Hmm they should be the same...but there could be an edge case

yoni206 (approved):

PR #134, opened 2020-12-17
Refactor tests to avoid code duplication

makaimann:
Uses a macro for adding tests in `CMake` and introduces a file for creating common transition systems.


lonsing (approved):
Hi @makaimann, looks good to me!

PR #135, opened 2020-12-17
Functional unroller implementation

makaimann:
This PR would add a functional unroller. This means it takes advantage of a functional transition system and can do substitution for unrolling. Because of this, it behaves slightly differently from a regular unroller. In particular, it will throw an exception if the term to be unrolled contains next state variables. Although it could just give the mapping from the next time-step, I thought it would be best to have this behavior because we don't actually want to directly unroll the transition relation. Then we would just get a bunch of trivial equalities, e.g. `functional_unroller.at_time(x.next, 4)` is structurally equal to `functional_unroller.at_time(x, 5)`.

Additionally, the `FunctionalUnroller` takes an `interval` parameter. If this is non-zero, it will create fresh symbols for the state variables every `interval` steps. This is to avoid exploding `ITE`s from pure functional unrolling. However, this also requires extra equality constraints between the fresh symbol and its semantics. These are provided with the function `extra_constraints_at(<time>)`. 

Let me know what you think! The main uses for this would be a functional BMC / k-induction implementation and for refinement in syntax-guided IC3.


makaimann:
I'm leaving this as a draft PR for now because it contains code from https://github.com/upscale-project/pono/pull/134 that I would like to merge to master first.

makaimann:
https://github.com/upscale-project/pono/pull/134 was merged so I promoted this to a full PR.

core/functional_unroller.cpp
```diff
@@ -0,0 +1,101 @@
...
+      }
+
+      Term fun_subst =
+          solver_->substitute(state_updates.at(v), time_cache_.at(t - 1));
+
+      if (create_new) {
+        // add equality to extra constraints
+        extra_constraints_[t] =
+            solver_->make_term(And,
+                               extra_constraints_[t],
```
ahmed-irfan, line 82: where it is populated for the first time?

engines/ceg_prophecy_arrays.cpp
```diff
@@ -93,6 +93,38 @@ CegProphecyArrays::CegProphecyArrays(const PonoOptions & opt,
 {
 }
 
+ProverResult CegProphecyArrays::prove()
```
ahmed-irfan, line 96: why is it in this PR?

ahmed-irfan (changes requested):
I see unnecessary changes. Can you try to rebase this branch with the master?

makaimann:
You were right. It needed `master` merged in.

engines/ceg_prophecy_arrays.cpp
makaimann: No longer changed after merging in master.

core/functional_unroller.cpp
makaimann: It starts as just `True` here: https://github.com/upscale-project/pono/blob/ca0490aa9b7137ccfcda31a8e57e7217c2952e77/core/functional_unroller.cpp#L52

core/functional_unroller.h
```diff
@@ -0,0 +1,89 @@
...
+#include "core/unroller.h"
+
+namespace pono {
+class FunctionalUnroller : public Unroller
+{
+ public:
+  /** Instantiate a FunctionalUnroller
+   *  @param ts the transition system to unroll
+   *  @param solver the solver to use
+   *  @param interval -- the interval to introduce new timed variables for state
```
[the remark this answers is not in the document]
makaimann: Hmm ran `clang-format` on this file and it didn't change anything. You're thinking for line length?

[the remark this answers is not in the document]
ahmed-irfan: i was not able to completely understand the text. I guess, you wanted some line breaks in them. That is why I asked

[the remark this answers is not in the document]
makaimann: Oh I see! Yeah I think that was because of `clang-format` actually. Fixed in https://github.com/upscale-project/pono/pull/135/commits/35ff261eba8eeab6fdc2317b2f650dec84712843

ahmed-irfan (approved):
LGTM. 

Thanks for taking into account the suggestions.

makaimann:
Of course, thanks for the detailed review!

PR #138, opened 2020-12-18
Use sequence interpolants in ic3ia refinement

makaimann:


makaimann:
The cluster is busy now but I'll submit jobs to evaluate this change soon.

makaimann:
Here are the results from the cluster run. The `*seq-interp*` directories are from this branch. Looks like it's better. Still a ways off from the `msat-ic3ia` backend, but it's a significant step in the right direction.

seq-interp-results

ahmed-irfan (approved):
LGTM.

PR #139, opened 2020-12-20
Move COI analysis to main

makaimann:
This PR moves cone of influence analysis to the main method. This way, the reduction happens before the transition system is even passed to an engine. The primary motivation for this change is that the engine might change the transition system (e.g. through abstraction) before running. This caused issues with cone of influence analysis that happened in `initialize`. This change was the simplest way we thought of for supporting cone-of-influence with engines like `ic3ia`. Furthermore, it seems like a nice logical separation in that it occurs completely separately from the engine.

### DIFF
--- a/engines/ic3ia.cpp
+++ b/engines/ic3ia.cpp
@@ -192,12 +192,6 @@ void IC3IA::check_ts() const
   // no restrictions except that interpolants must be supported
   // instead of checking explicitly, just let the interpolator throw an
   // exception better than maintaining in two places
-
-  if (options_.static_coi_) {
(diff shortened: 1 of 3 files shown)


ahmed-irfan (approved):
Tests are failing!

makaimann:
It’s just bison failing to download. Had issues with that earlier today also. Restarted the CI runs

lonsing (approved):
Looks good to me! This way the implementation is much cleaner.

makaimann:
Thanks for your reviews!

PR #141, opened 2020-12-21
Add basic IC3SA implementation

makaimann:
This PR would add basic support for syntax-guided IC3 from [Model Checking of Verilog RTL Using IC3 with Syntax-Guided Abstraction](https://link.springer.com/chapter/10.1007/978-3-030-20652-9_11) by Aman Goel and Karem Sakallah. This is only an initial implementation and is not very performant yet. Improvements can be made in subsequent PRs. In particular, there are still a lot of TODOs, but it is technically functional.


makaimann:
@ahmed-irfan just FYI, working on `IC3SA` here. It's not based just off master, so master will need to be merged once other PRs are in. I don't think this needs a review but feel free to make changes if you want.

makaimann:
I'm marking this ready for review, but also keeping it `do not merge` because the `Prover` interface will change soon. Once https://github.com/upscale-project/pono/pull/164 and associated PRs are merged, I'll update this branch. As mentioned in the description, this is not very performant yet. However, I thought it would be best to get the initial version merged and make incremental improvements from there.

makaimann:
Here are the results from the cluster runs:
ic3sa-results

I'm doing 20m runs now for more throughput, so they don't solve as many. Like I mentioned above, the performance is not very good but I think it can be improved separately. It's partially this implementation of ic3sa, but also partially the main IC3 implementation, which I am currently improving in a separate branch. I'm not worried about the 8 errors -- it looks like they are segfaults in the mathsat to string function for large terms. Although this brings up a related issue that the logger will call the `to_string` of a term, even if the verbosity is not high enough to print it. I can look into that more later.

makaimann:
Just FYI, made the changes to keep this in line with the recent prover updates. Started new jobs on the cluster to make sure nothing went wrong.

makaimann:
Still working after latest update.

Screenshot from 2021-01-06 11-33-51

ahmed-irfan:
> Still working after latest update.
> 
> Screenshot from 2021-01-06 11-33-51

It seems that the ic3ia got improved

makaimann:
> > Still working after latest update.
> > Screenshot from 2021-01-06 11-33-51
> 
> It seems that the ic3ia got improved

Yeah my guess is that it's related to the `reduce_predicates` unroller issue fix that we discussed offline.

engines/ic3sa.cpp
```diff
@@ -0,0 +1,880 @@
...
+      BVOr,
+      BVAnd });
+
+// main IC3SA implementation
+
+IC3SA::IC3SA(const Property & p,
+             const TransitionSystem & ts,
+             const smt::SmtSolver & solver,
+             PonoOptions opt)
+    : super(p, RelationalTransitionSystem(solver), solver, opt),
```
ahmed-irfan, line 57: why relational here?

makaimann: Good question. It's cause `IC3SA` learns path axioms during refinement that need to be added to the transition system, and those axioms might not be functional. I left a comment about it (way further down in the code), but I should mention it here too.

engines/ic3sa.cpp
```diff
@@ -0,0 +1,880 @@
...
+  // first populate with predicates
+  for (const auto & p : predset_) {
+    if (solver_->get_value(p) == solver_true_) {
+      cube_lits.insert(p);
+    } else {
+      cube_lits.insert(solver_->make_term(Not, p));
+    }
+  }
+
+  EquivalenceClasses ec = get_equivalence_classes_from_model(ts_.statevars());
```
ahmed-irfan, line 77: we are relying on the compiler that it doesn't create a new object and copies the elements. We should make sure that it is the case.

engines/ic3sa.cpp
```diff
@@ -0,0 +1,880 @@
...
+  construct_partition(ec, cube_lits);
+  IC3Formula cube =
+      ic3formula_conjunction(TermVec(cube_lits.begin(), cube_lits.end()));
```
ahmed-irfan, line 80: ditto

engines/ic3sa.cpp
makaimann: Gotcha, like pass it as a set instead?

engines/ic3sa.cpp
makaimann: This one we should check into but I'd rather not change the interface of `ic3formula_conjunction` right now because it's used everywhere. I'm pretty sure that return value optimization will pick this up.

engines/ic3sa.cpp
ahmed-irfan: yeah

engines/ic3sa.cpp
ahmed-irfan: good point! let's check it later

engines/ic3sa.cpp
makaimann: That didn't make sense -- I meant pass it as a reference to get updated?

ahmed-irfan: pass it as a reference and update in the method

engines/ic3sa.cpp
```diff
@@ -0,0 +1,880 @@
...
+  }
+
+  for (const auto & sv : ts_.statevars()) {
+    SortKind sk = sv->get_sort()->get_sort_kind();
+    if (sk != BOOL && sk != BV)
+    {
+      throw PonoException("IC3SA currently only supports bit-vectors");
+    }
+  }
+  for (const auto & iv : ts_.inputvars()) {
```
ahmed-irfan, line 194: inputs are already checked here https://github.com/upscale-project/pono/pull/141/files#diff-d8b374ffc4535468eca246f9584183ee392cc728e6fd3a86332eeec176369206R176

why are we checking them again now?

engines/ic3sa.cpp
ahmed-irfan: i see later in the code that it is needed.

engines/ic3sa.cpp
```diff
@@ -0,0 +1,880 @@
...
+  assert(!solver_context_);
+  assert(cex_.size());
+  // This function will unroll the counterexample trace functionally one step at
+  // a time
+  // it will introduce fresh symbols for input variables and will keep
+  // track of old model values to plug into inputs if an axiom is learned
+  Result r;
+  UnorderedTermMap last_model_vals;
+
+  UnorderedTermSet inputvars = ts_.inputvars();
```
ahmed-irfan, line 298: ts_ or conc_ts_?

ts_ doesn't have inputs. Right?

makaimann: Good catch! Clarified with 4a83034f8ae88f3ea63160900d98c4b9abbb7668.

engines/ic3sa.cpp
makaimann: Good catch, fixed with b659b910a5913771d31f3fe6b0eb2b6b5041e109

engines/ic3sa.cpp
makaimann: Done with 6296864291b91773f118cfc5f19574921297eceb.

engines/ic3sa.cpp
```diff
@@ -0,0 +1,875 @@
...
+        last = term;
+
+        // TODO: see if a DisjointSet would make this easier
+        // update representative for this class
+        if (!found_repr) {
+          if (term->is_symbolic_const()) {
+            repr = term;
+            repr_val = false;
+            found_repr = true;
+          } else if (!term->is_value() && repr_val) {
```
ahmed-irfan, line 666: is the else case possible?

soltelem.second contains variables, right?

engines/ic3sa.cpp
```diff
@@ -0,0 +1,875 @@
...
+      representatives.push_back(repr);
+    }
+
+    assert(representatives.size() == sortelem.second.size());
+
+    // add disequalities between each pair of representatives from
+    // different equivalent classes
+    for (size_t i = 0; i < representatives.size(); ++i) {
+      for (size_t j = i + 1; j < representatives.size(); ++j) {
+        const Term & ti = representatives.at(i);
```
ahmed-irfan, line 683: it can move outside of the inner loop

engines/ic3sa.cpp
```diff
@@ -0,0 +1,875 @@
...
+        const Term & tj = representatives.at(j);
+        // should never get the same representative term from different classes
+        assert(ti != tj);
+        if (ti->is_value() && tj->is_value()) {
```
ahmed-irfan, line 687: is it possible ?

engines/ic3sa.cpp
```diff
@@ -0,0 +1,875 @@
...
+          // if there are two values, don't need to express disequality
+          continue;
+        }
+        lit = solver_->make_term(Distinct, ti, tj);
+        if (!lit->is_value()) {
```
ahmed-irfan, line 692: again, is it possible to not have this?

engines/ic3sa.cpp
```diff
@@ -0,0 +1,876 @@
...
+  logger.log(
+      2,
+      "IC3SA::generalize_predecessor projecting on {}/{} state variables",
+      coi_symbols.size(),
+      ts_.statevars().size());
+
+
+  UnorderedTermSet cube_lits;
+  // first populate with predicates
+  for (const auto & p : predset_) {
```
ahmed-irfan, line 157: are predicates also going to be added to the cube? 
If so, why?

engines/ic3sa.cpp
makaimann: It actually contains interpreted constants also. The term abstraction includes everything. This is actually important cause you might need to learn that `x = 2` or something. It also guarantees completeness I think.

engines/ic3sa.cpp
makaimann: Yeah it is possible. For example, say you have `1` and `2` in your term abstraction, but in this particular model, none of the other terms are equal to them. Then they're the only representative of their class.

engines/ic3sa.cpp
makaimann: This one I'm not sure. It's possible we could remove it but sometimes it was adding `true`. Even if they aren't values, it might reduce to true through simplification e.g. `x + 2 == (x + 1 + 1)` would reduce to true.

engines/ic3sa.cpp
makaimann: Yeah the predicates are also added to the cube -- it's part of the abstraction as described in the paper. From a theoretical perspective I'm not sure if it's strictly necessary. It might or might not be. But it at least makes it more precise. Unnecessary ones should get removed by predecessor and inductive generalization anyway.

engines/ic3sa.cpp
makaimann: Good idea! Addressed in 9b08cd271c904995985b4cfa77cbf70e817a83c9

engines/ic3sa.cpp
ahmed-irfan: predicates can be other than just equality. Right? 
Then, the cubes could contain inequalities (coming from constraints), which is not just equality abstraction.

ahmed-irfan (approved):
I have questions more about the technique. Maybe, we can discuss it offline. 

For the PR, LGTM

makaimann:
Sounds good, let’s have a call tomorrow

engines/ic3sa.cpp
makaimann: Yeah that’s right. I think that’s why it’s called syntax guided instead of equality abstraction. It’s the predicate assignments plus equalities of other terms.

makaimann:
Thanks for the review! Going to merge now.

PR #142, opened 2020-12-21
Add reset solver feature to IC3Base

makaimann:
Adds option to reset solver after certain number of check sat calls (default is 5000).

### DIFF
--- a/engines/ic3base.cpp
+++ b/engines/ic3base.cpp
@@ -47,7 +47,8 @@ IC3Base::IC3Base(Property & p, SolverEnum se)
     : super(p, se),
       reducer_(create_solver(se)),
       solver_context_(0),
-      num_check_sat_since_reset_(0)
+      num_check_sat_since_reset_(0),
+      failed_to_reset_solver_(false)
 {
 }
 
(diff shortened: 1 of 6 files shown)


makaimann:
@ahmed-irfan here's the reset solver option we discussed before. I haven't tested it on the cluster yet so I marked it do not merge for now. Might be useful for your experiments though. Since it's related to clause subsumption, we can do that in this same PR if you think that makes sense.

engines/ic3base.cpp
```diff
@@ -694,6 +716,57 @@ void IC3Base::pop_solver_context()
   solver_context_--;
 }
 
+Result IC3Base::check_sat()
```
ahmed-irfan, line 719:
```suggestion
inline Result IC3Base::check_sat()
```

engines/ic3base.cpp
```diff
@@ -694,6 +716,57 @@ void IC3Base::pop_solver_context()
...
+{
+  num_check_sat_since_reset_++;
+  return solver_->check_sat();
+}
+
+Result IC3Base::check_sat_assuming(const smt::TermVec & assumps)
```
ahmed-irfan, line 725:
```suggestion
inline Result IC3Base::check_sat_assuming(const smt::TermVec & assumps)
```

ahmed-irfan (approved):
LGTM, only two minor comments

ahmed-irfan:
the PR is labeled as `do no merge`. @makaimann Do you plan to add something more in this PR?

makaimann:
Nope, it's ready! Just removed the label. Originally I planned to evaluate it on the cluster before merging but I don't think that's necessary anymore -- especially because it's an option. Instead, we can merge this and https://github.com/upscale-project/pono/pull/146#pullrequestreview-557264063 and evaluate them together.

PR #143, opened 2020-12-21
adds is_blocked method

ahmed-irfan:


engines/ic3base.cpp
```diff
@@ -394,7 +394,12 @@ bool IC3Base::rel_ind_check(size_t i,
 bool IC3Base::block_all()
 {
   while (has_proof_goals()) {
-    ProofGoal pg = get_next_proof_goal();
+    const ProofGoal &pg = get_next_proof_goal();
+    if (is_blocked(pg)) {
+      logger.log(3, "Skipping Already blocked proof goal <{}, {}>",
```
makaimann, line 399:
```suggestion
      logger.log(3, "Skipping already blocked proof goal <{}, {}>",
```

makaimann (approved):
Thanks for the PR! So it looks good to me in terms of correctness. I'm trying to understand how we get into this situation though where `is_blocked` will be true.

makaimann:
Just FYI I think this is good to go. I'll let you merge it in case you decide to tweak anything or whatnot.

PR #144, opened 2020-12-22
syguspdr

zhanghongce:
WIP

(I don't know if I should put partial model generation in `modifier` or not. Because I saw there is `coi` there, and I though this can be regarded as dynamic coi)


zhanghongce:
sorry, maybe I should first use rebase or something to remove those commits that you probably have squeezed into one...
(because this was based on the ic3 refactoring branch)

Let me close this first, and try some fixes locally. But please let me know if there is an exact command to do this. Thanks!

makaimann:
Hey @zhanghongce, thanks for starting this! You can try rebasing on master but that might give you lots of conflicts. Unless the rebase goes through easily, I'd recommend just merging master.

As for where to put partial model generation: that's a great question. I'm not sure it belongs in modifiers. Modifiers are supposed to be anything that changes the transition system, which static-coi does. I'd recommend either creating a new file in `utils`, or adding it to `https://github.com/upscale-project/pono/blob/master/utils/ts_analysis.h`. There's also `term_analysis.*` but I think that this type of partial model generation takes the specifics of the transition system into account, right?

zhanghongce:
Hi Makai
Thanks! I solve the rebase issue.
Actually its input is just a term (and assuming the solver was in the "SAT" status after a previous check). So it does not use any information of the transition system.

Issue #73, opened 2020-08-26
Handle state variables without next-function as inputs


lonsing:
Issue refers to BTOR2 inputs and Pono master 5b6af14f7b904d2c2140d4a8aab61dc2472506bb.

Excerpt from the BTOR2 paper (https://link.springer.com/content/pdf/10.1007%2F978-3-319-96145-3_32.pdf):

"A state variable without associated next function is treated as a primary input, i.e., it has the same behaviour as inputs defined via keyword _input_."

Pono properly handles inputs that are explicitly declared using the keyword _input_. Such inputs are added to the set `inputvars_` of a transition system.

However, state variables declared using the keyword _state_ and for which no next-function is given are added to the set `statevars_` of a transition system although according to the BTOR2 standard they should appear in `inputvars_`.

PR #145, opened 2020-12-22
WIP: SyGuS PDR

zhanghongce:


zhanghongce:
Please let me know if it is okay to put partial model generation under modifiers. Thanks!

makaimann:
Hey! Sorry, I responded just now on your closed PR. Copy-pasted that response here:

> As for where to put partial model generation: that's a great question. I'm not sure it belongs in modifiers. Modifiers are supposed to be anything that changes the transition system, which static-coi does. I'd recommend either creating a new file in utils, or adding it to https://github.com/upscale-project/pono/blob/master/utils/ts_analysis.h. There's also term_analysis.* but I think that this type of partial model generation takes the specifics of the transition system into account, right?

What do you think about putting it in `ts_analysis.*`? That's supposed to be for any functions that do "analysis" loosely defined on a transition system but don't change the transition system. It also makes perfect sense to just create new files there, especially if it's a lot of code.

zhanghongce:
No problem. Will move to `ts_analysis`.

makaimann:
Cool, yeah I think `term_analysis.*` or new files makes sense! Whichever you prefer.

zhanghongce:
Thanks!

makaimann:
Actually just looking through the diff, it's a fair amount of code. `term_analysis.*` is more for small utility functions in my opinion. How about you just move `partial_model.*` to the `utils` directory and otherwise leave it the same.

zhanghongce:
> Actually just looking through the diff, it's a fair amount of code. `term_analysis.*` is more for small utility functions in my opinion. How about you just move `partial_model.*` to the `utils` directory and otherwise leave it the same?

No problem. Thanks!

zhanghongce:
This update relies on the UNSAT core reduction function in this pull request: https://github.com/makaimann/smt-switch/pull/152

zhanghongce:
### Usage

`pono -e sygus-pdr  --no-assume-prop <input.btor>`


### Known Limitations

* It does not handle arrays
* Option `--no-assume-prop` is needed
* When there are assumptions in the input, the generated invariants may contain input variables (see the discussion in the end)
* When running on the same input multiple times, the number of frames it takes to converge might vary. I attribute this to the randomness in the fixedpoint of the UNSAT core reduction.
* It will check and see if there are state variables without next functions. If so, it will refuse to start. This is because in my original implementation, I also create primed variables for the input variables. When porting this to Pono, I made some changes to adapt for this difference, but this does not include the case for handling implicit inputs (state variables without next functions). Ideally, I can create a pass to auto-convert these implicit inputs to explicit input variables, but this will change `ts.cpp` and `ts.h` and may create more conflicts in merging. If later #73 is resolved, I'm happy to test and adapt it for that change.

Other issues found in testing:

* Sometimes there could be issues when using Boolector (for example when translating terms with uninterpreted functions, where translation is needed in unsatcore reducer), but you can switch to CVC4
* Sometimes, CVC4 seems to get stuck at `term->to_string()` but Boolector is fine

I only tested this port with Boolector and CVC4, I'm happy to fix any issues if there are when running with other SMT solvers.

### A discussion on the assumptions/constraints in the problem description

For example, in the following transition system, if there is an assumption that `s != 3` (i.e., s3 is never reached). Then it implicitly means `s == 2 => i==1`. So when there are assumptions/constraints in the problem description, I actually allow the algorithm to learn facts about both the state variables and the input variables.


```
     *     *      i=0
  s0 -> s1 -> s2 ----> s3
  ^            |i=1    |*
  |            |       |
  ----------------------

``` 

However, this will violate the assertions in `check_invar` function, which requires invariant to be only on the state variables.

So, when there are assumptions in the problem description, using `--check-invar` option may fail.

### Samples Added

I added a few btor2 samples in `samples/sygus-pdr`. Some of them may be difficult for other engines. 

I will continue to test it on more samples

makaimann:
Thanks @zhanghongce for working on this and the great description! I'll take a closer look at it tomorrow. 

Just a quick first impression related to the known limitations: what do you think about requiring that there be no input variables instead of no state variables without a next state? It would be easy to write a pass which changes inputs into state variables without next state updates and this would handle both that limitation and the `check_invar` issue. I don't like the idea of allowing lemmas over input variables unless we have a pass that converts them to "implicit" inputs. Due to the way `ts.next` functions with inputs, you could easily become unsound if there are lemmas over inputs.

zhanghongce:
> what do you think about requiring that there be no input variables instead of no state variables without a next state?

I think that would work. Thank you!

engines/syguspdr.cpp
```diff
@@ -0,0 +1,1542 @@
...
+  const TermVec &ac = a.children;
+  const TermVec &bc = b.children;
+  // NOTE: IC3Formula children are sorted on construction
+  //       Uses unique id of term, from term->get_id()
+  return ac.size() <= bc.size()
+         && std::includes(bc.begin(), bc.end(), ac.begin(), ac.end());
+}
+
+
+void SygusPdr::constrain_frame(size_t i, const IC3Formula & constraint,
```
makaimann, line 1383: Does this need to be overridden? What's the difference?

engines/syguspdr.cpp
```diff
@@ -0,0 +1,1542 @@
...
+  assert(proof_goals.empty());
+  return true;
+} // block_all
+
+// -----------------------------------------------------------------
+// allow input in the constraint (when there are additional constraints)
+// -----------------------------------------------------------------
+
+
+static bool subsumes(const IC3Formula &a, const IC3Formula &b)
```
makaimann, line 1370: If `constrain_frame` isn't overridden, this doesn't need to be either right? I'd like to move towards that -- it's better not to duplicate.

engines/syguspdr.cpp
```diff
@@ -0,0 +1,1542 @@
...
+
+  constrain_frame_label(i, constraint);
+  frames_.at(i).push_back(constraint);
+}
+
+// -----------------------------------------------------------------
+// Require the capability to reset solver
+// -----------------------------------------------------------------
+
+void SygusPdr::reset_solver()
```
makaimann, line 1419: This looks the same as the base class implementation so it shouldn't be overridden. If you need to do extra work in `reset_solver` I think the best thing to do is override, and use `super::reset_solver` to do the basic parts.

makaimann: btw, I'll refactor that soon so that adding the label constraints is separate from resetting the solver. I'll implement that tomorrow and let you know.

engines/syguspdr.cpp
```diff
@@ -0,0 +1,1542 @@
...
+  for (size_t j = i; j < frames_.size(); ++j) {
+    for (const auto &u : frames_[j]) {
+      res = solver_->make_term(And, res, u.term);
+    }
+  }
+  return res;
+}
+
+// Difference : I need it to call the new constrain_frame
+// which allow input in the constraints
```
makaimann, line 1481: This can be fixed with the pass that converts to state variables. I'm just commenting here as a reminder to remove that after the pass is implemented.

engines/syguspdr.cpp
```diff
@@ -0,0 +1,1542 @@
...
+    }
+  }
+
+  // get rid of garbage at end of frame
+  Fi.resize(k);
+
+  return Fi.empty();
+} // propagate
+
+// difference: lift the restriction on only_curr vars
```
makaimann, line 1517: This will be fixed by the pass that converts to state variables.

engines/syguspdr.cpp
```diff
@@ -0,0 +1,1542 @@
...
+
+
+// ------------------------------------------------------------------
+//  May Block 
+
+/**
+ * Priority queue of proof obligations inspired by open-source ic3ia
+ * implementation
+ */
+class MayProofGoalQueue
```
makaimann, line 861: This is the same, right? If you need access to it here, we can just expose it in the base class. But I'd prefer to just make changes to the base class so that you don't need access.

engines/syguspdr.h
```diff
@@ -0,0 +1,213 @@
...
+#include "modifiers/op_abstractor.h"
+#include "utils/sygus_ic3formula_helper.h"
+#include "utils/partial_model.h"
+#include "utils/syntax_analysis_common.h"
+#include "utils/syntax_analysis_walker.h"
+#include "utils/syntax_analysis.h"
+
+namespace pono {
+
+// this class is added simply because SygusPdr would like to see such a transition system
```
makaimann, line 38: This can be removed after the pass that converts all inputs to state variables.

engines/syguspdr.h
```diff
@@ -0,0 +1,213 @@
...
+  std::unordered_map<smt::Term, syntax_analysis::IC3FormulaModel *> model2cube_;
+  //std::unordered_map<syntax_analysis::IC3FormulaModel *, syntax_analysis::IC3FormulaModel *>
+  //  to_full_model_map_;
+  syntax_analysis::cex_term_map_t cex_term_map_;
+
+  PartialModelGen partial_model_getter;
+
+  IC3Formula get_initial_bad_model();
+  std::pair<IC3Formula, syntax_analysis::IC3FormulaModel *>
+    ExtractPartialModel(const smt::Term & p);
```
makaimann, line 175: For functions could we use the _ syntax?

engines/syguspdr.h
```diff
@@ -0,0 +1,213 @@
...
+  std::pair<IC3Formula, syntax_analysis::IC3FormulaModel *>
```
makaimann, line 176: ditto

engines/syguspdr.h
```diff
@@ -0,0 +1,213 @@
...
+    ExtractInitPrimeModel(const smt::Term & p_prime);
```
makaimann, line 177: ditto

engines/syguspdr.h
```diff
@@ -0,0 +1,213 @@
...
+  std::tuple<IC3Formula, syntax_analysis::IC3FormulaModel *, syntax_analysis::IC3FormulaModel *>
```
makaimann, line 178: ditto

engines/syguspdr.h
```diff
@@ -0,0 +1,213 @@
...
+    ExtractPartialAndFullModel(syntax_analysis::IC3FormulaModel * post_model);
+
+  std::unique_ptr<syntax_analysis::OpExtractor> op_extract_;
+
+  syntax_analysis::TermScore term_score_walker_;
+  unsigned GetScore(const smt::Term & t);
```
makaimann, line 184: ditto -- underscore syntax

utils/syntax_analysis_common.h
```diff
@@ -0,0 +1,279 @@
...
+  }; // class BvConstructs
+
+  class SyntaxStructure{
+
+  public:
+    typedef std::unordered_map<uint64_t, BvConstructs> SyntaxT;
+
+    bool new_constructs;
+
+    const SyntaxT & GetSyntaxConstruct() const {
```
makaimann, line 240: Could you use the underscore syntax for functions?

utils/syntax_analysis_common.h
```diff
@@ -0,0 +1,279 @@
...
+    void insert_const    (uint64_t width, const std::string & name);
+    void insert_op_unary (uint64_t width, smt::PrimOp);
+    void insert_op_binary(uint64_t width, smt::PrimOp);
+    void insert_op_comp  (uint64_t width, smt::PrimOp);
+    void insert_concat   (uint64_t width, concat_t  && );
+    void insert_extract  (uint64_t width, extract_t && );
+    void insert_rotate   (uint64_t width, rotate_t  && );
+    void insert_extend   (uint64_t width, extend_t  && );
+
+    void CutVars(
```
makaimann, line 253: ditto wrt the underscore syntax for functions for this set of functions.

utils/syntax_analysis_walker.h
```diff
@@ -0,0 +1,267 @@
...
+
+namespace pono {
+namespace syntax_analysis {
+
+class Walker {
+protected:
+  // if you want to buffer and avoid further walk
+  // Skip is your chance to do that, return true
+  // if already buffered
+  virtual bool Skip(const smt::Term & ast) = 0;
```
makaimann, line 32: ditto for underscore syntax + the rest of the functions in this file

makaimann (changes requested):
Hey @zhanghongce, thank you for working on this! It looks really great! I made a first pass and left quite a few comments but most of them are minor. The big picture feedback from the review is:
1. I'd like to stick to the syntax rules: Capital case for classes, lowercase and underscores for functions
2. It looks like you made a lot of effort to avoid changing the base class. I'd much prefer to just make the necessary updates and avoid duplicating code. I think a pass that converts inputs to state variables could solve a lot of the issues you were circumventing. That plus a few more tweaks and you probably don't need to override very many functions. In general, I prefer not to override unless absolutely necessary, but I'd like to *never* shadow a method. So if you must override it and it's not virtual, let's just make it virtual.

I can become more involved to help with these final changes.

makaimann:
Additionally, with respect to learning lemmas over input variables: I'm really interested in discussing this with you further. I'm still somewhat opposed to that idea. Although for SyGuS PDR it makes sense to just make everything a state variable IMO. 

In case you're curious, here's my reasoning for being opposed to inputs in lemmas. The goal is to build an inductive invariant. That invariant should describe a set of states. Inputs are by-definition not part of the state representation. Inputs should *only* be involved in transitions and an inductive invariant says nothing about transitions. In the example system you gave above, you wouldn't need to learn a lemma over `i` because the system would already have `s != 3 /\ s' != 3`. Thus you could never get that transition from `s2` to `s3`. This does imply that `s2 -> i=1` but you don't need to learn that as a lemma because it's already entailed by the transition relation.

Anyway, I just thought I would put my thoughts here for reference. But, I'd really like to discuss this more (either here in the PR or in a Zoom call later). I'm open to being convinced otherwise! :)

zhanghongce:
Thanks for the feedback. I also don’t want the duplicate code (with just minor changes) there. However, there are certain cases that prohibits SyGuS PDR to smoothly fit in right now and I don’t know how much I can bother you to adapt for it. Also the underlying code base has been constantly changing in the recent week, what is there is what I believe to be working with SyGuS PDR. In the future, if there is a reliably steady interface, I also want to remove the duplicates. Let me elaborate some of my points below:

-------------------


First, when it is possible, I really want to avoid doing many special casing. It makes the code messier and most importantly, more error-prone. Take the state-input issue as an example. There are now three cases:

* normal state var: has primed version, has next function
* input var: no primed version, no next function
* implicit input vars: has primed version, but no next function

This matters for the predecessor generalization I’m using, because I need to replace the variables in a formula in the next step with their update functions and then do the partial model extraction. For those without update functions I want to turn them into a variable that won’t mess up with the current step (which is their primed version in my own implementation). In my own implementation, every variable (input and state) has a primed version and I also keep a “prime_var→next_function”  map. So this can be done simply by: 

`substitute( ts_.next(m),   prime_var→next_function )`

Now I need to deal with the 3 cases and mostly importantly, reason to my self that those special case handling that I wrote will keep the rest of algorithm intact.

In general, I’m okay to limit the lemmas to only on the “normal state variables”. However, for the primed variable issue, I think keeping a prime version of variables for inputs will make my life much easier (I can keep this outside the class `TransitionSystem`, and this is actually part of the reason why there is a `CustomTransitionSystem` class in `sygus_pdr.h`. The other part of the reason is: I want to use `to_next_func` in `FunctionalTransitionSystem` which is not public, but I guess I can do this also outside the class but with more function calls, however, a recent change makes `ts_` not a pointer anymore, which makes it a lot harder for me to use this `CustomTransitionSystem` with original base class functions and this forms part of the reason of duplicate code) 

---

It is a similar issue in the handling of property in a frame.

Thanks for taking my advice of changing the `intersects_bad` with `reaches_bad`. But, I also feel that there are a lot of special casing here. For example, `trans_label_` needs to imply the property and `get_frame_term` also need to take account into the property and so on. 

In my original implementation, I just put the property as a lemma (IC3Formula) in F[1]. And lemma propagation will ensures it appears in every frame that needs it (it is certainly push-able, ensured by the fact that `reaches_bad` already finds no states violating that before pushing stage) and no need for other special handling in other parts of the code, but I guess this will break your assumption for bit-level IC3? Since there is no guarantee that this property is a bit-level clause?


---

There are many changes going on in the underlying code base, and some of them, though small, have caused a lot of mysterious failures after merging and it took a lot of efforts to pin-point the locations. 

One of the example is adding the property to part of the transition relation (I mean the use of `prop_in_trans` function)  and another example is that `rel_ind_check` suddenly starts to do some work which I think should be the work of `inductive_generalization` in a recent change (I’m referring to the unsat core extraction in `rel_ind_check`). If the interface from base class is reliably stable (I mean not only the type signature but also the expectations of what exactly goes in and out), then I believe there won’t be the need for a lot of the shadowing functions. 

---
With that said, there might be one exception, it is the function for blocking may-block goals.  May-block goals are different from must-block goals, and some of the assertions there (e.g., `assert(i < 2 || !check_intersects(out.term, get_frame_term(i – 2)))` in `rel_ind_check` does not fit) and some of the handling will be slightly different (e.g., no need for requeue), so I think keeping it as a separate one might be clearer.

---

All in all, I understand there must be a balance and trade-off in the base class for all these different flavors of IC3 and my initial intention was to not bother you with too many demands from my side, and this resulted in the current redundancy in the code. It will be greatly helpful if any adaptation can be made there. Below I attached a list of changes I wish could happen (but no pressure, here). Thanks!

1. Keeping `ts_` as a pointer can help make a lot of the base class code re-useable, so I don’ t need to shadow them. Alternatively, it will be great if there could be a pass that converts all input variables to state variables without next functions so they will have their primed versions.
2. For `rel_ind_check`, I expect it to be just a relative inductive check as its name suggested, specifically, I hope that I can expect what `get_model_ic3formula` or `predecessor_generalization` generate will be exact the same one that feeds a later `inductive_generalization` and used in the SMT checking in `rel_ind_check`. (In other words, no need for unsat core reduction in `rel_ind_check` because it un-do some of the work done by my `inductive_generalization`)
3. if possible, treat the property as a IC3formula in the frame rather than using trans_label_→property and other special handling. If this change is hard to make, I think I can also make adaptation for this, but I really hope the choice on this will be stable, so that I don’t need to worry about re-work it.
4. No change needed, but I hope in the future there will still be the option of *not* making property as part of the transition relation, otherwise I will have to figure out how to adapt for it in the future.
5. Could you please keep the order of `get_model_ic3formula` and `predecessor_generalization` stable? Previously this was “either-or” and then it became that `get_model_ic3formula` needs to feed the `predecessor_generalization` ’s input and this underlying structure change has taken a lot of efforts for me to re-work, so I hope whichever you choose, this will be stable.

Thank you for reading this! I'm happy for further discussion on any part of it.

makaimann:
Thanks for the summary! So yes, I recognize that it was difficult to keep up with the upstream changes. I was trying to optimize the base implementation to improve performance of other IC3 flavors which resulted in several non-trivial changes. It is currently stable, and I have no plans to change it except for adapting it to better support `SyGuS-PDR`. So from now on, you can assume that `IC3Base` will stay the same except for changes we make for this PR.

Regarding the specific changes you mentioned:
1. Pass is already in a PR https://github.com/upscale-project/pono/pull/214. Should be merged soon. I prefer this to making ts a pointer.
2. Maybe we can discuss this elsewhere? I see why you might want it to stay the same, but I don't see how it undoes any work of `inductive_generalization`. An unsat core is a cheap generalization that `inductive_generalization` can just build on. I think I'll need to better understand what SyGuS-PDR expects to figure out how to best incorporate this. It makes other flavors of IC3 noticeably faster because it cheaply generates clauses and re-generates when pushing to higher frames. We could consider making it a separate method and / or optional but there's always a tradeoff in complexity
3. I'll think about this more. I'd also prefer to keep the property in the frame, but as you mentioned it doesn't work well with versions of IC3 where the property isn't guaranteed to fit the restrictions of an `IC3Formula`. One thing we could do is always introduce a monitor for the property. What do you think about that?
4. If 3 is handled then it doesn't need to be in the transition relation anymore. What is the issue here out of curiosity? You can't assume the property in the pre-state?
5. See point above -- IC3Base is now stable except for changes related to this PR.

makaimann:
> I don’t know how much I can bother you to adapt for it.

Thanks for the thought! Not a bother at all -- you can ask me to adapt it as much as needed. :) I think it worked great for you to just do whatever was needed to get it working, and sorry that it was difficult with the upstream changes. But now, we can work together to put the finishing touches on it! I'm happy to make changes, although I think it would help to discuss some things in more detail on a call if you're free. Like you observed, we need to balance the different variants.

zhanghongce:
Thanks a lot for your reply!

1. Thank you very much! That will be a lot of help for me.
2. Yes. It might work better but it might also cause problems as sometimes it may over-generalize a proof goal (which still holds for the bounded trace explored so far, but if lemma generation is based on that goal, the generated lemma might be myopic). And the whole idea of SyGuS-PDR is to hand this over to the SyGuS procedure to decide how to generalize.

Let me give an example. 

```
module cnt(clk, out);
   parameter WIDTH=4;

   input                    clk;
   output wire [WIDTH-1:0]   out;

   reg [WIDTH-1:0] cnt1 = 0 ;
   reg [WIDTH-1:0] cnt2 = 4'b1111;

   always @(posedge clk) begin
      cnt1 <= cnt1 + 1;
      cnt2 <= cnt2 - 1;
   end

   assign out = cnt1 & (4'b1111-cnt2);

   assert property ( !(cnt1 == 4'b0110 && cnt2 == 4'b0110) );

endmodule
```

We know that the assertion holds because of an inductive invariant: `cnt1 ==  4'b1111-cnt2`.

However, the bad state like (`cnt1 == 4'b0110 && cnt2 == 4'b0110`) can be generalized to only `cnt1 == 4'b0110` in the first several frames because it either cannot be reached yet. And when you give this to the inductive generalization procedure, the generated lemmas will be only around `cnt1` and therefore, got misled.

3. No need of extra work from your side. I can adapt. (I think mostly this might not be a problem, because I can choose to take the transition relation out for any assertion I need and not use the labels any way)

4. It might also be okay to have the property in the transition relations. I’m just not confident about this, and I need to check in my code and reason about it. I think as long as transition relation is always used in conjunction with some frame, then whether the property appears in a frame or in the transition relation will anyway results in the same expression. But there is also code in the SyGuS part that runs something like a simulation using this transition relation (and no frame there at all), where I need to think carefully how it affects.

Thank you very much for your help! Especially PR #214. Greatly appreciated.

makaimann:
Cool, thanks for the great explanation on number 2! That makes a lot of sense to me. I misunderstood and didn't realize you were worried about overgeneralization. And that's a great problem to demonstrate the issue with myopic generalization. I was aware of that issue from an intuitive point of view (and from the gspacer paper), but the example definitely helps concretize it. I'd still like to keep that as an option, because it does help other IC3 variants quite a bit. For IC3IA my guess is that's because the abstraction-refinement primarily handles relevancy, and then unnecessary predicates can (usually) be pruned out of the lemma. Although I'm sure there are examples where this could cause issues anyway. I'm leaning towards making it an option now that can just be disabled for SyGuS-PDR. You also make a good point about it not necessarily being the role of `rel_ind_check`. However, I prefer to keep the solver contexts consistent within a function whenever possible (i.e. I try not to pop a context that was pushed in another function because that gets really confusing). So to use the unsat core it needs to be done during the check. I'll make a PR with the option on it and get your review before merging it if you don't mind.

For 3 and 4. I'm definitely happy to work on this regardless. It seems like as long as you're only focused on the first property violation, you should be able to assume the property in any previous state. That being said, I also think it would be cleaner to keep the property in the frame instead of a global label assumption, so I'll look into this more. Additionally, I wasn't so sure about making `assume-prop` (in the pre-state) a default option. It does help with some abstraction techniques, but it does feel strange. So if it causes problems with SyGuS-PDR, I'll just turn it off by default.

zhanghongce:
Thanks, Makai! Another question regarding the constraints just occurred to me. For the following transition system, if the assumption is `s != 7` and the assertion is `s != 4` . Is the expected outcome to be safe or unsafe? I mean if you just look at the first 5 steps, then s4 is reachable. But the extension of that in the future time will violate the assumption `s != 7`.


```
                 i=0
s0 --> s1 --> s2 --> s3 --> s4 --> ... --> s7
 ^             | i=1                       |
 |             |                           |
 -------------------------------------------
```

makaimann:
Good question! So I think it should be unsafe. Here's my thought process: I'm thinking in terms of paths and that system has a path which reaches a property violation without violating any assumptions. Another way to say that is that once you add the assumption `s != 7` you've created this transition system instead:
```
                 i=0
s0 --> s1 --> s2 --> s3 --> s4 --> ... --> s6
 ^             | i=1                       |
 |             |                           |
 -------------------------------------------
```

which deadlocks at `s6`. At least that's my feeling on it. Plus BMC would say unsafe so I think it's good to be consistent with that. I believe our existing IC3 variants would also say unsafe but I should check haha.

Any extra thoughts @ahmed-irfan?

makaimann:
Also, a quick thought on `to_next_func`. With #214, you don't need to worry about inputs anymore. And `to_next_func` is actually just (internal) syntactic sugar for `solver_->substitute(term, ts.state_updates())`. Would you be able to use that instead? If not, I can definitely make `to_next_func` public. Note: it replaces the _current_ state version (not next state) of variables with their next state update, and it will do nothing for implicit inputs. If you need something else, please let me know and I'll add it to `TransitionSystem`.

makaimann:
I just merged master into this branch and resolved the conflicts in 81c11b30bd792756c07a8e8015029cc1e2e51894. Hope you don't mind. It was pretty straightforward so I don't think I messed anything up.

zhanghongce:
> Also, a quick thought on `to_next_func`. With #214, you don't need to worry about inputs anymore. And `to_next_func` is actually just (internal) syntactic sugar for `solver_->substitute(term, ts.state_updates())`. Would you be able to use that instead? If not, I can definitely make `to_next_func` public. Note: it replaces the _current_ state version (not next state) of variables with their next state update, and it will do nothing for implicit inputs. If you need something else, please let me know and I'll add it to `TransitionSystem`.

Yes, I think you don't need to make it public. Thanks!


> I just merged master into this branch and resolved the conflicts in 81c11b3. Hope you don't mind. It was pretty straightforward so I don't think I messed anything up.

Okay, I will work from that (I also have some local commits not pushed yet, and I will solve the conflicts locally if there is any)
Thanks!

makaimann:
> Okay, I will work from that (I also have some local commits not pushed yet, and I will solve the conflicts locally if there is any)

Cool! And feel free to revert 81c11b30bd792756c07a8e8015029cc1e2e51894 and/or force push if needed.

makaimann (approved):
Hi @zhanghongce, thanks for all your work on this -- it's much appreciated. I just made another pass and I think it looks awesome! It looks really clean.

makaimann:
@zhanghongce let me know if you'd like to make any other changes, otherwise I think it looks great and I'm ready to merge it!

p.s. I decided not to worry about the function case for now. It's not that important, and we can always change it later if at all.

zhanghongce:
Hi Makai, I think code-wise it is okay. I have been running larger tests and working on locating and debugging some performance issues. The new commits solves those that I found, but I think there is still room for more.I can definitely improve more in the future. Thanks!

makaimann:
Gotcha, that sounds great! Let's get this merged then and you can continue to work on it (but of course no pressure). Either way I think it's good to get this in and improve on it in subsequent PRs. Thanks!

I'll wait for Travis and then merge.

zhanghongce:
Thank you, Makai! And the updated usage is `pono --promote-inputvars -e sygus-pdr <btor name>`

zhanghongce:
Hi Makai, I'm looking at the results from Travis and it shows segmentation faults on two existing tests:

```
The following tests FAILED:
	  7 - test_witness (SEGFAULT)
	 24 - test_btor2_ts_copy_equal (SEGFAULT)
```

However, I don't think I touch the base class and I'm not able to reproduce this segmentation fault by `make test` locally. If you have any clues that would be really helpful. Thanks!

makaimann:
Interesting! I'll pull and see if I can reproduce locally.

makaimann:
> I'm not able to reproduce this segmentation fault by `make test` locally.
I'm not able to reproduce it locally either. I'll keep looking into it, but it's very strange.

makaimann:
Actually, I have an idea what this could be -- I was updating CVC4 for smt-switch and it probably got pulled into this PR. Hopefully it's fixed now. I'll restart the Travis build.

makaimann:
All right, looks like that was it. Going to merge now! Thanks again for all your work!!

zhanghongce:
> All right, looks like that was it. Going to merge now! Thanks again for all your work!!

Thank you very much!

PR #146, opened 2020-12-22
Ic3 subsumption check

ahmed-irfan:

PR #147, opened 2020-12-22
ic3* code -- minor changes

ahmed-irfan:


engines/ic3.cpp
```diff
@@ -51,7 +51,7 @@ IC3Formula IC3::get_model_ic3formula(TermVec * out_inputs,
   const UnorderedTermSet & statevars = ts_->statevars();
   TermVec children;
   children.reserve(statevars.size());
-  for (auto sv : ts_->statevars()) {
+  for (auto const &sv : ts_->statevars()) {
```
makaimann, line 54: I have a slight preference for `const auto &`. They're semantically the same but since elsewhere we write `const <type-specifier> &`, I think it's more consistent.

engines/ic3.cpp
```diff
@@ -69,7 +69,7 @@ IC3Formula IC3::get_model_ic3formula(TermVec * out_inputs,
   }
 
   if (out_inputs) {
-    for (auto iv : ts_->inputvars()) {
+    for (auto const &iv : ts_->inputvars()) {
```
makaimann, line 72: ditto

makaimann (approved):
This looks great! Thanks for making these improvements -- much cleaner / more efficient. I left a minor comment with a personal preference. I didn't comment on all the relevant spots because there are a lot. Feel free to make the change or not, up to you. Otherwise, I think it's good to go except for a merge conflict after the latest PR got merged.

ahmed-irfan:
Please also take a look at the changes in the propagate method

makaimann:
Yeah I saw that, looks good to me. It should be semantically equivalent, right? You're just editing the vector in place by overriding the first `k` elements and then resizing -- very nice. As long as `constrain_frame` is after popping the solver context (which it is) I think it's all good.

After merging https://github.com/upscale-project/pono/pull/146 then `constrain_frame` in this `propagate` should be called with `false`, right?

engines/ic3.cpp
ahmed-irfan: done in 4a22594

engines/ic3.cpp
ahmed-irfan: done in 4a22594

ahmed-irfan:
@makaimann can you please check the last two commits and merge if everything looks fine.

PR #149, opened 2020-12-23
Minor fix for msat-ic3ia backend

makaimann:


ahmed-irfan (approved):
Lgtm 

thanks for catching this.

was this the only place ?

makaimann:
Yep, that was the only spot. After this fix it compiles on my machine.

PR #151, opened 2020-12-23
Output number of initial predicates with logger

makaimann:


engines/ic3ia.cpp
```diff
@@ -202,10 +202,14 @@ void IC3IA::initialize()
...
   get_predicates(solver_, ts_->init(), preds, false);
+  size_t num_init_preds = preds.size();
   get_predicates(solver_, bad_, preds, false);
+  size_t num_prop_preds = preds.size() - num_init_preds;
   for (const auto &p : preds) {
     add_predicate(p);
   }
+  logger.log(1, "Number predicates found in init: {}", num_init_preds);
+  logger.log(1, "Number predicates found in prop: {}", num_prop_preds);
   // more predicates will be added during refinement
```
ahmed-irfan, line 213: perhaps, let's also print the total number of predicates we add initially.

ahmed-irfan (approved):
LGTM.

Only one minor comment.

engines/ic3ia.cpp
makaimann: Cool, added with 4c4251db7f7e4effea18f6b0bff5287bc4765316

PR #152, opened 2020-12-23
Create method is_predicate

makaimann:
Factored out `is_predicate` check so that it can be used in other places.


ahmed-irfan (approved):
LGTM

PR #154, opened 2020-12-27
simplify prover interface

ahmed-irfan:


makaimann (approved):
Looks really good, thanks for working on this! It's much simpler. I saw that you also moved the code setting the default solver options from the `Prover` constructor to `create_solver` which makes more sense.

makaimann:
FYI @zhanghongce only one constructor needed for `Prover` implementation now.

zhanghongce:
Got it. Thanks!

> FYI @zhanghongce only one constructor needed for `Prover` implementation now.

PR #156, opened 2020-12-29
Use priority queue to manage proof goals

makaimann:
Use a priority queue for managing proof goals -- which will always return a proof goal at the lowest frame available. This PR also fixes an issue where proof goals were popped too early. Instead, they should only be removed from the queue when it has been blocked.

### DIFF
--- a/engines/ic3base.cpp
+++ b/engines/ic3base.cpp
@@ -658,16 +658,6 @@ void IC3Base::assert_trans_label() const
   solver_->assert_formula(trans_label_);
 }
 
-bool IC3Base::has_proof_goals() const { return !proof_goals_.empty(); }
-
-ProofGoal * IC3Base::get_next_proof_goal()
-{
-  assert(has_proof_goals());
-  ProofGoal * pg = proof_goals_.top();
-  proof_goals_.pop();
-  return pg;
-}
-
 void IC3Base::add_proof_goal(const IC3Formula & c, size_t i, ProofGoal * n)
 {
   // IC3Formula aligned with frame so proof goal should be negated
--- a/engines/ic3base.cpp
+++ b/engines/ic3base.cpp
@@ -41,6 +41,43 @@ static bool term_hash_lt(const smt::Term & t0, const smt::Term & t1)
   return (t0->hash() < t1->hash());
 }
 
+/** Checks if a contains all the terms of b
+ *  NOTE: can't rely on std::includes because it uses operator<
+ *  and smt-switch doesn't have a meaningful operator< on terms
+ *  i.e. can't be used for equality (term_hash_lt wouldn't work)
+ *  @param a vector of terms
+ *  @param b vector of terms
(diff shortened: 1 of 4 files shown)


makaimann:
Marked `do not merge` for now while we're evaluating it.

makaimann:
FYI @zhanghongce, once this PR is merged it will change the interface slightly -- it uses `ProofGoal *` instead of just `ProofGoal`

engines/ic3base.cpp
```diff
@@ -129,10 +129,11 @@ ProverResult IC3Base::check_until(int k)
   int i = reached_k_ + 1;
   assert(i >= 0);
   while (i <= k) {
-    // reset cex_pg_ to null
+    // reset cex_pg_->to null
     // there might be multiple abstract traces if there's a derived class
     // doing abstraction refinement
-    cex_pg_ = ProofGoal();
+    cex_pg_ = new ProofGoal();
```
ahmed-irfan, line 135: where do we  delete ?

engines/ic3base.cpp
```diff
@@ -301,7 +302,7 @@ ProverResult IC3Base::step_0()
   Result r = check_sat();
   if (r.is_sat()) {
     const IC3Formula &c = get_model_ic3formula();
-    cex_pg_ = ProofGoal(c, 0, nullptr);
+    cex_pg_ = new ProofGoal(c, 0, nullptr);
```
ahmed-irfan, line 305: ditto

engines/ic3base.cpp
```diff
@@ -398,16 +399,24 @@ bool IC3Base::block_all()
       reset_solver();
     }
 
-    const ProofGoal &pg = get_next_proof_goal();
+    ProofGoal * pg = get_top_proof_goal();
```
ahmed-irfan, line 402: are we going to change this later? if not then it should be a pointer to const object

ahmed-irfan:
left some comments

engines/ic3base.cpp
makaimann: Oops, good catch. Thanks! Fixed with 5c927cdc9fbcad0b3f270f0ba117cd447278a72a.

engines/ic3base.cpp
makaimann: Thanks! Fixed with 5c927cdc9fbcad0b3f270f0ba117cd447278a72a

engines/ic3base.cpp
makaimann: Good point! Updated with 9f493ef01f129dbc798555683b27a762a3bf5b98

makaimann:
Looks like performance actually got worse. It still seems like this is the right thing to do though. What do you think @ahmed-irfan?

Screenshot from 2020-12-29 10-03-36

ahmed-irfan (approved):
LGTM.

I think that we are doing the right changes in this PR. We should merge it.

PR #158, opened 2020-12-29
Separate COI analysis from TS transformation

makaimann:
There are uses for COI analysis that don't result in transforming the TS. This PR would just separate the analysis functions into a separate class which is used by the TS transforming `StaticConeOfInfluence` class.


utils/fcoi.cpp
```diff
@@ -154,12 +193,12 @@ void ConeOfInfluence::compute_term_coi(const Term & term,
           assert(ts_.inputvars().find(cur) == ts_.inputvars().end());
           assert(!ts_.is_next_var(cur));
           logger.log(3, "collect COI statevar {}", cur);
-          collect_coi_term(new_coi_state_vars, cur);
+          new_coi_state_vars.insert(cur);
```
lonsing, line 196: Good catch to get rid of the redundant check whether newly inserted elements already appear in the set! I was not aware of that implicit feature of unordered sets---too much C programming in the past, where you have to take care of all that yourself. :-)

lonsing (approved):
Looks good to me, nice feature!

PR #160, opened 2020-12-29
Add option to replace init and prop with fresh state variables

makaimann:
As discussed offline, this option can help ic3ia significantly by reducing the number of initial predicates. Sometimes it is good to seed with predicates from init/prop, but when there are hundreds of predicates in them (which can happen with the BTOR2 format) then it will overwhelm the procedure with too many unnecessary predicates.


makaimann:
Here are the results from the cluster run. We're still not quite as good as the open-source `ic3ia` built directly on mathsat, but this option makes a big difference:
Screenshot from 2020-12-29 10-12-36

makaimann:
FYI this option currently doesn't work with `--static-coi` because removed state variables are left in `init` which causes problems when we add constraints over `init` later. This seems like more of an issue on the `coi` side, so I don't think it needs to be handled in this PR, except maybe an exception when trying to use both. What do you think?

ahmed-irfan:
> FYI this option currently doesn't work with `--static-coi` because removed state variables are left in `init` which causes problems when we add constraints over `init` later. This seems like more of an issue on the `coi` side, so I don't think it needs to be handled in this PR, except maybe an exception when trying to use both. What do you think?

Yeah, it is ok for now to throw an exception when both options are used.

ahmed-irfan:
oh, another thought.

How about we apply init-prop-transformation first and then apply static-coi on the resulting TS? 
Will it work?

makaimann:
We could do that, the only problem then is that all the initial state constraints will be in the `constraints` instead of `init`, and so `COI` will be very limited cause it'll be much more likely to keep variables (any variable in the same constraint is kept in COI analysis). However, that's probably still better than it not working at all, so I'll do that instead of throwing an exception. I'll also do a conjunctive partition of init to limit the impact.

makaimann:
Done with 769200996a262a8dbe7c00bea73e4bbdaf2e3535 and a337dc403139d7351291f33c722ec7ffd94e6284

ahmed-irfan (approved):
LGTM.

PR #163, opened 2020-12-31
Prover update 1

ahmed-irfan:
first PR of the series.

It simplifies the prover class and also the inherited classes

### DIFF
--- a/core/prop.h
+++ b/core/prop.h
@@ -27,18 +27,11 @@ class Property
   Property(const TransitionSystem & ts, const smt::Term & p,
            std::string name="");
 
-  /** Copies property to a new solver
-   *  @param prop the property to copy
-   *  @param tt the term translator to use
-   *  @return a property using the solver in tt
-   */
-  Property(Property & prop, smt::TermTranslator & tt);
-
   ~Property();
 
   const smt::Term prop() const { return prop_; }
 
-  TransitionSystem & transition_system() { return ts_; }
+  const TransitionSystem & transition_system() const { return ts_; }
 
   std::string name() { return name_; };
 
--- a/engines/kinduction.cpp
+++ b/engines/kinduction.cpp
@@ -21,9 +21,10 @@ using namespace smt;
 
 namespace pono {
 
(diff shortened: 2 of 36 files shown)


core/prop.h
```diff
@@ -27,18 +27,11 @@ class Property
...
-   *  @return a property using the solver in tt
-   */
-  Property(Property & prop, smt::TermTranslator & tt);
-
   ~Property();
 
   const smt::Term prop() const { return prop_; }
 
-  TransitionSystem & transition_system() { return ts_; }
+  const TransitionSystem & transition_system() const { return ts_; }
```
makaimann, line 34: Should we still include the transition system at all? I was thinking it could just be the solver, name, and the property term.

engines/kinduction.cpp
```diff
@@ -109,14 +112,14 @@ bool KInduction::inductive_step(int i)
...
   Term disj = false_;
-  for (auto v : ts_->statevars()) {
-    Term vi = unroller_.at_time(v, i);
-    Term vj = unroller_.at_time(v, j);
-    Term eq = solver_->make_term(PrimOp::Equal, vi, vj);
-    Term neq = solver_->make_term(PrimOp::Not, eq);
+  for (const auto &v : ts_.statevars()) {
+    const Term &vi = unroller_.at_time(v, i);
+    const Term &vj = unroller_.at_time(v, j);
+    const Term &eq = solver_->make_term(PrimOp::Equal, vi, vj);
```
makaimann, line 121: Can this be a `const` reference to a term that was just created? It seems like the object would be destroyed.

makaimann (commented):
Thanks for making these changes, looks like a lot of work! Overall looks pretty good to me, but I left a few comments. Let me know if I can help with the python bindings.

core/prop.h
makaimann: I see -- it's in the second part of this PR.

ahmed-irfan:
closing this because #164 will be merged directly.

PR #164, opened 2020-12-31
Prover update 2

ahmed-irfan:
2nd PR

merge this after #163


ahmed-irfan:
@makaimann adding monitor code in the prover class was giving trouble in ceg-prophecy because the concrete and abstract-ts would not be in sync if we do monitor addition in prover. 

So, I thought to move into the modifiers dir and do the monitor addition same as we do coi

engines/kinduction.cpp
```diff
@@ -109,14 +112,14 @@ bool KInduction::inductive_step(int i)
...
   Term disj = false_;
-  for (auto v : ts_->statevars()) {
-    Term vi = unroller_.at_time(v, i);
-    Term vj = unroller_.at_time(v, j);
-    Term eq = solver_->make_term(PrimOp::Equal, vi, vj);
-    Term neq = solver_->make_term(PrimOp::Not, eq);
+  for (const auto &v : ts_.statevars()) {
+    const Term &vi = unroller_.at_time(v, i);
+    const Term &vj = unroller_.at_time(v, j);
+    const Term &eq = solver_->make_term(PrimOp::Equal, vi, vj);
```
makaimann, line 121: Mentioned this in part 1, but reproducing here. Is it a problem to have a `const` reference to a term that was just created? Won't the object be destroyed?

modifiers/prop_monitor.h
```diff
@@ -0,0 +1,60 @@
+/*********************                                                  */
+/*! \file mod_init_prop.h
```
makaimann, line 2: update filename

pono.cpp
```diff
@@ -287,9 +293,15 @@ int main(int argc, char ** argv)
         StaticConeOfInfluence coi(fts, { prop }, pono_options.verbosity_);
       }
 
+      if (!fts.only_curr(prop)) {
+        logger.log(1, "Got next state or input variables in property. "
+                   "Generating a monitor state.");
+        prop = add_prop_monitor(fts, prop);
```
makaimann, line 299: I'm wondering if we should put this in the `Prover`. I'm thinking for the API so people don't need to know they should call this before passing to a prover.

makaimann (commented):
It's looking good to me! I just left a few comments. Let me know if I can help with the python bindings. I'll make a second pass once Travis passes.

engines/kinduction.cpp
ahmed-irfan: done in 1f37892

pono.cpp
ahmed-irfan: yeah, that is a problem. But, if we do it in the prover, we have the problem to update different ts (orig_ts, ts, conc_ts) at different levels of the prover class inheritence.  I didn't find a easy solution to that. That is why I did it outside of the prover.  It is not perfect solution. Maybe we should revisit it in the future.

engines/prover.cpp
```diff
@@ -50,8 +52,14 @@ void Prover::initialize()
...
   reached_k_ = -1;
-  bad_ = solver_->make_term(smt::PrimOp::Not, property_.prop());
-  assert(ts_->only_curr(bad_));
+
+  if (!bad_) {
+    // initialize bad_ if it is not set already
+    const Term &prop_term = (ts_.solver() == orig_property_.solver())
+      ? orig_property_.prop()
+      : to_prover_solver_.transfer_term(orig_property_.prop());
+    bad_ = solver_->make_term(smt::PrimOp::Not, prop_term);
```
makaimann, line 61: Could we still have a check here that `bad_` only contains current state variables?

modifiers/prop_monitor.h
```diff
@@ -0,0 +1,60 @@
...
+**/
+
+#pragma once
+
+#include "core/ts.h"
+#include "utils/logger.h"
+
+namespace pono {
+
+smt::Term add_prop_monitor(TransitionSystem & ts, const smt::Term & prop)
```
makaimann, line 24: Could we expose this function in the Python API as well? It's useful for calling engines directly.

makaimann:
Left a couple more minor comments. And for Travis, looks like python bindings are working, but the python tests need to be updated.

engines/prover.cpp
ahmed-irfan: done in c03dd5d

modifiers/prop_monitor.h
ahmed-irfan: done in a149678

makaimann (approved):
Looks good to me, thanks for making these changes! Only minor comment is that in the `Prover`, you could consider throwing an exception if `bad_` is not only current variables, rather than asserting it. That's because currently the user is responsible for making sure that's the case (if using the API) so they'll need to know when something is wrong.

Also, should we just merge this directly? Or did you want to do https://github.com/upscale-project/pono/pull/163 first?

ahmed-irfan:
About merging this directly or doing #163 first, I actually don't have a preference. What do you think?

makaimann:
Let's just merge it directly! I think it's good to go if you're okay with it? I'll wait for Travis first though.

zhanghongce:
Hi I notice that this pull request change initialization of `ts_`. So if a derived class want to use an abstracted transition system to replace it, then it should use copy assignment to overwrite `ts_`, right?

PR #165, opened 2021-01-03
Prover update 3: adds Engine enum to Pover

ahmed-irfan:
first merge #163 and #164


engines/bmc.cpp
```diff
@@ -25,6 +25,7 @@ Bmc::Bmc(const Property & p, const TransitionSystem & ts,
          const SmtSolver & solver, PonoOptions opt)
   : super(p, ts, solver, opt)
 {
+  engine_ = Engine::BMC;
```
makaimann, line 28: What do you think about putting this in the initializer list instead? I have a slight preference for that just because it doesn't need to be in the body and I think it's more visible above.

makaimann (approved):
This looks great, thanks for making the change! Left a minor comment that you can address or not.

engines/bmc.cpp
ahmed-irfan: `engine_` is declared in prover (base class). As far as I think, if we want to set it in the intialization-list, we have to pass the engine enum in all the constructors.

That's is why I did this way.

makaimann: Ahh gotcha, makes sense!

PR #167, opened 2021-01-05
Prover update 4: unifies the engines interface

ahmed-irfan:


engines/bmc.cpp
```diff
@@ -25,6 +25,7 @@ Bmc::Bmc(const Property & p, const TransitionSystem & ts,
          const SmtSolver & solver, PonoOptions opt)
   : super(p, ts, solver, opt)
 {
+  engine_ = Engine::BMC;
```
makaimann, line 28: This isn't important at all, but I'm not sure why this diff is showing up. This line is already in `master`: https://github.com/upscale-project/pono/blob/f91ef4d6d8459a219712f1c3d120ed10b8cfd8be/engines/bmc.cpp#L28

makaimann (approved):
LGTM, thanks for making this update!

engines/bmc.cpp
ahmed-irfan: yeah, i am also curious. I tried to rebase, but it was giving conflicts, so I didn't do it

PR #168, opened 2021-01-05
Fixes for msat-ic3ia after prover-update

makaimann:
This PR just has some basic fixes for the `msat-ic3ia` backend (after https://github.com/upscale-project/pono/pull/164), which is currently not tested on Travis.


ahmed-irfan (approved):
Thanks. I missed those

PR #169, opened 2021-01-05
Set solver options and refactor smt solver related code

makaimann:
This PR changes the `smt_solver_` option to use `SolverEnum`s and adds support for `create_solver_for` which for certain solver/engine pairs will set good default options automatically. In particular, this PR sets options for MathSAT that are good for IC3 variants. Future PRs can set good default options for other solver/engine pairs.


ahmed-irfan (approved):
LGTM

PR #170, opened 2021-01-05
cegar and cegar-proph template

ahmed-irfan:
This PR makes the CEGAR and CEGAR-Proph into template classes.


ahmed-irfan:
@makaimann note that I changed the unroller in the prover.h to be an adaptive-unroller.

We can later make the adaptive unroller the default unroller and remove the old unroller.

makaimann (approved):
LGTM, thanks for the incremental PRs! Very basic question, why do you need to use `super::*`?

makaimann:
> @makaimann note that I changed the unroller in the prover.h to be an adaptive-unroller.
> 
> We can later make the adaptive unroller the default unroller and remove the old unroller.

Sounds good, I think it makes sense to make the default unroller "adaptive".

ahmed-irfan:
> LGTM, thanks for the incremental PRs! Very basic question, why do you need to use `super::*`?

It is a template thing. Notice that the data members are non-depdendent, so we can also do `this->` or `using`. 
What is your preference?

makaimann:
> It is a template thing. Notice that the data members are non-depdendent, so we can also do `this->` or `using`.
> What is your preference?

I don't feel strongly. Unless you prefer one of the others, I say sticking with `super::` is fine

PR #173, opened 2021-01-06
various: initialize in axiom-enumerator, abstractor: make do_abstraction public, update unroller, update ceg-proph code 

ahmed-irfan:
sorry, I went on doing various changes, but they are not huge.


makaimann:
Are you planning to remove the `unroller_` from the `Prover` in this PR or a separate one?

ahmed-irfan:
> Are you planning to remove the `unroller_` from the `Prover` in this PR or a separate one?

I am going to do it in a separate pr. Btw: I thought about it and it would not change much what I am doing in this pr

makaimann:
> I am going to do it in a separate pr. Btw: I thought about it and it would not change much what I am doing in this pr

Gotcha, sounds good! So is this ready for another review once the conflicts are resolved?

ahmed-irfan:
> > I am going to do it in a separate pr. Btw: I thought about it and it would not change much what I am doing in this pr
> 
> Gotcha, sounds good! So is this ready for another review once the conflicts are resolved?

Yes, it is ready for review

refiners/array_axiom_enumerator.cpp
```diff
@@ -142,6 +142,10 @@ ArrayAxiomEnumerator::ArrayAxiomEnumerator(ArrayAbstractor & aa,
 {
   conc_bad_ = solver_->make_term(Not, prop);
   false_ = solver_->make_term(false);
+}
+
+void ArrayAxiomEnumerator::initialize()
```
makaimann, line 147: What do you think about adding an `initialized_` flag that is checked in `enumerate_axioms`? It could be an assertion so it's only included in debug mode, but it's just an extra check that the developer didn't forget to initialize.

modifiers/implicit_predicate_abstractor.h
```diff
@@ -55,9 +55,9 @@ class ImplicitPredicateAbstractor : public Abstractor
                          const smt::TermVec & new_preds,
                          smt::TermVec & out);
 
- protected:
   void do_abstraction() override;
```
makaimann, line 58: What do you think about adding an `abstracted_` flag that's checked later to ensure that the abstraction was done before using the object? The checks could be in `assert` so they're removed in release mode. Made a similar comment for axiom enumerator.

makaimann (approved):
Looks good to me, left some minor comments.

refiners/array_axiom_enumerator.cpp
ahmed-irfan: good idea. I am going to do it now.

modifiers/implicit_predicate_abstractor.h
ahmed-irfan: good idea! going to change it now

engines/ceg_prophecy_arrays.cpp
```diff
@@ -158,9 +160,11 @@ void CegProphecyArrays<Prover_T>::initialize()
...
 {
+  aa_.do_abstraction();
+  aae_.initialize();
   // the ArrayAbstractor already abstracted the transition system on
   // construction -- only need to abstract bad
-  Term prop_term = (super::ts_.solver() == super::orig_property_.solver())
+  Term prop_term = (abs_ts_.solver() == super::orig_property_.solver())
                        ? super::orig_property_.prop()
                        : super::to_prover_solver_.transfer_term(
                            super::orig_property_.prop());
```
makaimann, line 170: minor thing I just noticed, this should probably be: `to_prover_solver_.transfer_term(super::orig_property_.prop(), BOOL)`. Basically that handles the case where the original solver was boolector and the property is a size-one bit-vector, then it will cast it to be a boolean if the current solver is not boolector.

refiners/array_axiom_enumerator.cpp
```diff
@@ -484,6 +433,74 @@ bool ArrayAxiomEnumerator::check_nonconsecutive_axioms(AxiomClass ac,
   return num_found_lemmas;
 }
 
+
+// protected methods
+
+void ArrayAxiomEnumerator::collect_arrays_and_indices()
```
makaimann, line 439: I assume this is all just moved, right?

makaimann (approved):
Looks good to me! I left one comment that I just noticed which is actually unrelated to this PR. If you'd rather get this merged, I'm fine with that and I can fix it separately.

refiners/array_axiom_enumerator.cpp
ahmed-irfan: yes

engines/ceg_prophecy_arrays.cpp
ahmed-irfan: similar situation might happen elsewhere. What do you think about creating a separate PR for that?

makaimann: Yep I'll take care of it.

PR #174, opened 2021-01-06
Refactor proof queue

makaimann:
This PR refactors the `IC3Base` code such that the queue of proof goals is a local variable instead of a member variable. This makes it much clearer when goals are added and removed.

This addresses the assertion failure that @zhanghongce brought up.

### DIFF
--- a/engines/ic3base.cpp
+++ b/engines/ic3base.cpp
@@ -171,8 +171,8 @@ ProverResult IC3Base::check_until(int k)
   ProverResult res;
   RefineResult ref_res;
   int i = reached_k_ + 1;
-  assert(i >= 0);
-  while (i <= k) {
+  assert(reached_k_ + 1 >= 0);
+  for (size_t i = reached_k_ + 1; i <= k; ++i) {
     // reset cex_pg_ to null
     // there might be multiple abstract traces if there's a derived class
     // doing abstraction refinement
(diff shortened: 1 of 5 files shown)


makaimann:
Just FYI, I know this isn't the cleanest. I'm planning to make several improvements to `IC3Base` (I've been working on debugging performance in a branch and will make a sequence of PRs based on what I found there). So, this will likely get refactored again, but I wanted to prioritize unblocking @zhanghongce. This should work for now, there's just a lot happening in one function -- `block_all`. I'd appreciate your feedback on this PR as well @zhanghongce.

I'm currently running debug cluster runs to check for any errors.

engines/ic3base.cpp
```diff
@@ -399,98 +390,112 @@ bool IC3Base::rel_ind_check(size_t i,
...
+          // TODO might have to change this if there's an algorithm
+          // that refines but can keep proof goals around
+          proof_goals.clear();
+
+          // and reset cex_pg_
+          if (cex_pg_) {
+            delete cex_pg_;
+            cex_pg_ = nullptr;
+          }
+          continue;
```
ahmed-irfan, line 422: this is not needed

makaimann: The `continue`? Actually it is because otherwise the rest of the loop body will run on an old proof goal.

makaimann: But that's one of the things I can refactor later to make more clear

ahmed-irfan: right. I think I didn't fully see the indentation

ahmed-irfan (approved):
LGTM

makaimann:
Screenshot from 2021-01-06 15-34-02

Here's the cluster run. Lost a few benchmarks, but I'm not that worried about it because it was a debug run. I'll merge this now since it fixes a bug, and refactor it more later. The errors are model evaluation errors in mathsat.

Additionally, just FYI, the diff in this PR makes it look worse than it is. It's really just moving code around.

zhanghongce:
Thanks a lot! Will merge to my branch and test and see.

PR #175, opened 2021-01-07
Refactor inductive generalization

makaimann:
Only return a single `IC3Formula` instead of a vector from `inductive_generalization`. It only used to return a `vector` before because of the interpolation mode for `ModelBasedIC3`. In that case, there was no guarantee that the interpolant was a single clause and could actually be a conjunction of formulae. However, that algorithm is not very good anyway and it complicated the rest of the code. In particular, there was an issue when trying to push the learned `IC3Formula` to higher frames. It should have been considering the whole vector at once, but instead it was checking each one individually. It's much simpler to just use a single `IC3Formula`. The `ModelBasedIC3` interpolation mode still works, it just might return an `IC3Formula` with only one child. I.e. the interpolant was a conjunction instead of a disjunction so it's just returned as a single formula.

### DIFF
--- a/engines/ic3.cpp
+++ b/engines/ic3.cpp
@@ -93,10 +93,9 @@ bool IC3::ic3formula_check_valid(const IC3Formula & u) const
   return true;
 }
 
-std::vector<IC3Formula> IC3::inductive_generalization(size_t i,
-                                                      const IC3Formula & c)
+IC3Formula IC3::inductive_generalization(size_t i, const IC3Formula & c)
 {
-  assert(!c.is_disjunction());  // expecting a cube
+  assert(!c.disjunction);  // expecting a cube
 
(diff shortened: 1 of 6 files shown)


makaimann:
Screenshot from 2021-01-06 17-51-41

Here is the cluster run showing there's no introduced bugs.

makaimann:
FYI @zhanghongce , with this PR `inductive_generalization` would only return a single `IC3Formula`. I believe that should work fine for you, but let me know if not.

ahmed-irfan (approved):
LGTM

zhanghongce:
> 
> 
> FYI @zhanghongce , with this PR `inductive_generalization` would only return a single `IC3Formula`. I believe that should work fine for you, but let me know if not.

No problem here. Thanks!

PR #177, opened 2021-01-07
Improvements and default implementation of inductive_generalization

makaimann:
This PR gives a default implementation of `inductive_generalization` and `generalize_predecessor`. The default is that `generalize_predecessor` just gives the model, but `inductive_generalization` aggressively tries to drop conjuncts and also re-uses `rel_ind_check` which can further generalize using an unsat core.


makaimann:
I'm running on the cluster to check for errors and evaluate performance. Marked `do not merge` until those runs are done.

engines/ic3base.cpp
```diff
@@ -346,32 +432,69 @@ bool IC3Base::rel_ind_check(size_t i, const IC3Formula & c, IC3Formula & out)
...
+
+  Result r = check_sat_assuming(assumps_);
   if (r.is_sat()) {
-    if (options_.ic3_pregen_) {
-      out = generalize_predecessor(i, c);
-    } else {
-      out = get_model_ic3formula();
+    if (get_pred) {
+      if (options_.ic3_pregen_) {
+        out = generalize_predecessor(i, c);
```
ahmed-irfan, line 457: why not pass `out` as a parameter to `generalize_predecessor`?

engines/ic3base.cpp
```diff
@@ -346,32 +432,69 @@ bool IC3Base::rel_ind_check(size_t i, const IC3Formula & c, IC3Formula & out)
...
+      } else {
+        out = get_model_ic3formula();
```
ahmed-irfan, line 459: `generalize_predecessor` is doing the same thing. No?

makaimann: Only by default, if it's overridden by the subclass then it might actually generalize.

engines/ic3base.h
```diff
@@ -273,6 +252,32 @@ class IC3Base : public Prover
    */
   virtual IC3Formula ic3formula_negate(const IC3Formula & u) const;
 
+  /** Attempt to generalize before blocking in frame i
```
ahmed-irfan, line 255: is something changed here? or is it just moved?

engines/ic3base.cpp
makaimann: Yeah I think that makes sense. So `get_model_ic3formula` is always called outside of it? The only reason that might not work well is that depending on the generalization approach (e.g. in a derived class) it might need more than just the cube. It might also ask for input / next state values, in which case it would call `get_model_ic3formula` again but with `out_inputs` and `out_nexts` non-null. But, I could refactor it more so that getting the input and next state variable values is a totally separate method. What do you think of that?

engines/ic3base.h
makaimann: Just moved!

engines/ic3base.cpp
ahmed-irfan: i see. then what about doing this `if (options_.ic3_pregen_)` check in generalize_predecessor when  we override in a derived class

makaimann: I don't like that as much because then every time you add a derived class you have to know to do that check. This way, it's only done once in the base class and you don't have to worry about it in the derived class.

ahmed-irfan: yeah. that is a good point.

makaimann:
Screenshot from 2021-01-07 13-08-56

The jobs on the cluster finished, looks pretty good to me.

The errors are all model evaluation errors (for some reason this happens often on the `cal*` family, particularly `cal220` and `cal234`) and one failure to find new predicates.

One thing to note, these jobs ran on b3eb4420edb76a3228c1d0d09323568e4d14bbfe but I don't think the changes since then are important enough for re-running.

ahmed-irfan (approved):

engines/ic3base.cpp
ahmed-irfan: i think, it is fine the way it is now.

ahmed-irfan:
> Screenshot from 2021-01-07 13-08-56
> 
> The jobs on the cluster finished, looks pretty good to me.
> 
> The errors are all model evaluation errors (for some reason this happens often on the `cal*` family, particularly `cal220` and `cal234`) and one failure to find new predicates.
> 
> One thing to note, these jobs ran on [b3eb442](https://github.com/upscale-project/pono/commit/b3eb4420edb76a3228c1d0d09323568e4d14bbfe) but I don't think the changes since then are important enough for re-running.

yeah, looks nice !

engines/ic3base.cpp
makaimann: Made those changes in 5c176ebd6046ac14f9bc0148c0282f5064ed3ec3. Let me know what you think.

ahmed-irfan: lgtm

makaimann:
FYI @zhanghongce change in interface. Renamed `generalize_predecessor` to `predecessor_generalization` and it takes a third argument which starts as the full assignment predecessor and then needs to be updated with the general version. Also separated the functions for getting input / next state values from the regular `get_model_ic3formula`.

zhanghongce:
> 
> 
> FYI @zhanghongce change in interface. Renamed `generalize_predecessor` to `predecessor_generalization` and it takes a third argument which starts as the full assignment predecessor and then needs to be updated with the generalized version. Also separated the functions for getting input / next state values from the regular `get_model_ic3formula`.

Got it. Thanks!

PR #178, opened 2021-01-07
Refactor and add assertions in ic3ia for reduce_predicates

makaimann:
This PR just refactors slightly and adds a few assertions and logger messages around the `reduce_predicates` phase of `IC3IA`.

### DIFF
--- a/engines/ic3ia.cpp
+++ b/engines/ic3ia.cpp
@@ -289,6 +289,10 @@ RefineResult IC3IA::refine()
   TermVec red_preds;
   if (ia_.reduce_predicates(cex, fresh_preds, red_preds)) {
     // reduction successful
+    logger.log(2,
+               "reduce predicates successful {}/{}",
+               red_preds.size(),
+               fresh_preds.size());
     if (red_preds.size() < fresh_preds.size()) {
       fresh_preds.clear();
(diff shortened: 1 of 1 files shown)


engines/ic3ia.cpp
```diff
@@ -134,10 +134,20 @@ void IC3IA::initialize()
...
+    // ia_ automatically includes all boolean variables
+    assert(p->is_symbolic_const() && p->get_sort() == boolsort);
+    predset_.insert(p);
+  }
+
   // add all the predicates from init and property to the abstraction
   // NOTE: abstract is called automatically in IC3Base initialize
   UnorderedTermSet preds;
-  get_predicates(solver_, ts_.init(), preds, false, false, true);
+  get_predicates(solver_, conc_ts_.init(), preds, false, false, true);
```
ahmed-irfan, line 150: essentially ts_.init() and conc_ts_.init() are the same. It is just more clearer to use conc_ts_.
RIght?

makaimann: Yeah that's right, I actually only changed it because I got confused about it myself :P But yes they should be the same.

ahmed-irfan (approved):

PR #179, opened 2021-01-07
Keep state variables occurring in INIT in COI analysis

makaimann:
As discussed elsewhere, there was an issue with state variables being removed from the system. Since COI ignores the initial state constraints, some removed variables may appear in `init`. When those are removed, the transition system becomes invalid (i.e. it contains non-variables in init). This prevented us from use COI with some techniques. This PR simply doesn't remove those state variables. This should not impact performance too much because those variables *only* occur in init. 

If it ends up affecting performance negatively, we can look into analyzing init to remove the state variables. This is not so simple though, because even for functional transition systems `init` can be arbitrary formulas over current state variables.

### DIFF
--- a/core/ts.cpp
+++ b/core/ts.cpp
@@ -476,7 +476,7 @@ void TransitionSystem::rebuild_trans_based_on_coi(
 {
   /* Clear current transition relation 'trans_'. */
   trans_ = solver_->make_term(true);
-  
+
   /* Add next-state functions for state variables in COI. */
   for (auto state_var : state_vars_in_coi) {
     Term next_func = NULL;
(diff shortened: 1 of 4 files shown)


makaimann:
FYI, since this is a simple change, I wasn't planning to run on the cluster. However, I'll be running lots of other jobs this week so if this introduces performance issues, I'll notice after it's merged.

lonsing (approved):
Thanks for that fix, looks good to me!

lonsing:
@makaimann, I agree with your comments on potential difficulties in analyzing `init` if it consist of arbitrary formulas. As you pointed out in a discussion elsewhere, it would be simple if `init` was a conjunction of equalities. Could we perhaps enforce that via the API? Or would that be too much of a restriction and limitation of usability?

makaimann:
> @makaimann, I agree with your comments on potential difficulties in analyzing `init` if it consist of arbitrary formulas. As you pointed out in a discussion elsewhere, it would be simple if `init` was a conjunction of equalities. Could we perhaps enforce that via the API? Or would that be too much of a restriction and limitation of usability?

That's a good question. I think a conjunction of equalities is the most common case but is too restrictive in general. One possible solution is to just have a flag that gets set if a more general constraint is added and then we can check to flag to see how much simplification we can do. What do you think about that? I was thinking we could address that later and just merge this PR for now though if that works for you?

lonsing:
Yes of course, sounds good, let's merge that PR.

PR #181, opened 2021-01-08
IC3Base: Use rel_ind_check in propate and find_highest_frame

makaimann:


makaimann:
Waiting on cluster run for performance comparison.

makaimann:
Screenshot from 2021-01-07 19-39-24

Runs finished. Only one more solve compared to current master but I think this change also simplifies the code.

ahmed-irfan (approved):
LGTM

engines/ic3base.cpp
```diff
@@ -641,42 +645,35 @@ bool IC3Base::is_blocked(const ProofGoal * pg)
...
+    assert(c.children.size());
+
+    // NOTE: rel_ind_check works on conjunctions
+    //       need to negate
+    if (rel_ind_check(i + 1, ic3formula_negate(c), gen, false)) {
+      // can push to next frame
+      // got unsat-core based generalization
+      assert(gen.term);
+      assert(gen.children.size());
+      constrain_frame(i + 1, ic3formula_negate(gen), false);
```
ahmed-irfan, line 668: If I understand correctly, we are doing generalization now -- we were not doing it before. RIght?

makaimann: Yeah that's right. Not the expensive minimizing generalization, but just the cheap unsat-core based generalization that happens automatically in `rel_ind_check`

PR #186, opened 2021-01-08
fix reset solver in ic3ia

ahmed-irfan:
I had to make some changes in implicit_abstractor and few other changes in the ic3ia code

### DIFF
--- a/engines/ic3ia.cpp
+++ b/engines/ic3ia.cpp
@@ -61,7 +61,7 @@ IC3IA::IC3IA(const Property & p,
 
 IC3Formula IC3IA::get_model_ic3formula() const
 {
-  const TermVec & preds = ia_.predicates();
+  const UnorderedTermSet & preds = ia_.predicates();
   TermVec conjuncts;
   conjuncts.reserve(preds.size());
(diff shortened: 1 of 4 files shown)


makaimann:
I'm not sure I follow this fix. In what situation would there be predicates in the abstraction that weren't added through the `ia_`? As long as it's added there then the `trans` is automatically updated and when you reset the solver the predicate abstraction bi-implications are re-added automatically (because they're in `trans`)

ahmed-irfan:
> I'm not sure I follow this fix. In what situation would there be predicates in the abstraction that weren't added through the `ia_`? As long as it's added there then the `trans` is automatically updated and when you reset the solver the predicate abstraction bi-implications are re-added automatically (because they're in `trans`). Before the reset, it only needs to add the assumption to `trans_label_` because that isn't updated when we updated `trans`.

Right! 
I guess this PR is a side-effect of working late ;-)

ahmed-irfan:
Not needed after clarification

makaimann:
I can add a comment somewhere that makes this clearer though -- it's definitely subtle.

PR #187, opened 2021-01-08
Incremental ceg-prophecy

ahmed-irfan:


makaimann (approved):
LGTM, thanks for the updates!

makaimann:
I'm starting cluster runs to evaluate this. Lets wait to merge it until we have those results.

makaimann:
Screen Shot 2021-01-10 at 2 33 07 PM

There was only one error (a segfault on `2019/wolf/2019A/picorv32_mutCY_mem-p1.btor`), so that doesn't seem too bad. However, there doesn't seem to be any performance increase which is strange. Plus this overall technique seems to be much worse than I would expect (although that might just be within the 20m timeout). I'll run msat-ic3ia for 20m as a comparison now.

makaimann:
Screenshot from 2021-01-11 09-48-46

Here's some new info. Looks like array benchmarks just take much longer than bv benchmarks in general (not surprising). These results compare the `msat-ic3ia` backend on array benchmarks with a 20m vs 1hr timeout -- big difference.

PR #188, opened 2021-01-08
Change implicit abstraction class

ahmed-irfan:
This PR removes the local copy of predicate maintained by implicit-abstractor. It also modifies the interface of the abstractor.


makaimann (approved):
LGTM! Much nicer to not have two copies of the predicates.

PR #190, opened 2021-01-11
Various minor fixes

makaimann:
This PR fixes a few minor issues in the IC3 variants backends

* sets the random seed in msat-ic3ia
* adds a fallback in case the bad cube can't be generalized
* adds an assertion in the `ic3ia` `get_model_ic3formula` that the value is actually a value
  * this came up because one of the mathsat options caused it to not get values for certain formulae

### DIFF
--- a/engines/ic3ia.cpp
+++ b/engines/ic3ia.cpp
@@ -63,12 +63,14 @@ IC3Formula IC3IA::get_model_ic3formula() const
 {
   TermVec conjuncts;
   conjuncts.reserve(predset_.size());
+  Term val;
   for (const auto &p : predset_) {
-    if (solver_->get_value(p) == solver_true_) {
+    if ((val = solver_->get_value(p)) == solver_true_) {
       conjuncts.push_back(p);
     } else {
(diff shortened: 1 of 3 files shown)


engines/ic3ia.cpp
```diff
@@ -63,12 +63,14 @@ IC3Formula IC3IA::get_model_ic3formula() const
 {
   TermVec conjuncts;
   conjuncts.reserve(predset_.size());
+  Term val;
   for (const auto &p : predset_) {
-    if (solver_->get_value(p) == solver_true_) {
+    if ((val = solver_->get_value(p)) == solver_true_) {
```
ahmed-irfan, line 68: just to be sure, what are you checking here?

makaimann: Yeah good question. This shouldn’t be an issue but it’s just a sanity check that the value is actually a value, like `true` or `false`. When I tried setting bool model generation only, it wouldn’t give a value for all terms, it would just return the same term.

ahmed-irfan (approved):

PR #191, opened 2021-01-12
Improved IC3IA options

makaimann:
Here are the changes we discussed offline that improve performance for IC3IA using mathsat.

Note: there are some downsides to this code. For one, there are problems with boolector that require the `base-context-1` option to work around. The issue there is that boolector will do top level propagation automatically on assertions at the top-level which causes issues (variables become arbitrary terms etc..). If you think it would be better, I could try a different solution which doesn't actually use predicate variables in the frames, but just for evaluating the model. The idea would be that all the frames still have the actual predicates in them, but when we check the values, we use the indicator variable so that we can use `bool_model_generation` with mathsat.

I'll give that a try and see if the code ends up being simpler. We probably don't have to wait to merge this if you don't want to, I can always submit another PR.

### DIFF
--- a/contrib/setup-smt-switch.sh
+++ b/contrib/setup-smt-switch.sh
@@ -3,7 +3,8 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 DEPS=$DIR/../deps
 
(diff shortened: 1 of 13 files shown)


makaimann:
Decided to close this in favor of https://github.com/upscale-project/pono/pull/192

PR #192, opened 2021-01-12
Improved IC3IA options

makaimann:
This PR gives some better options to mathsat for the IC3IA backend. Instead of using predicate variables though, this PR just uses indicator variables for evaluating a model. I think this is a better solution than https://github.com/upscale-project/pono/pull/191.

### DIFF
--- a/contrib/setup-smt-switch.sh
+++ b/contrib/setup-smt-switch.sh
@@ -3,8 +3,7 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 DEPS=$DIR/../deps
 
-# TEMP pull branch
-# SMT_SWITCH_VERSION=92371eb947f2641c161adf70d3d9be5fc60b9e75
+SMT_SWITCH_VERSION=9230477a5b9cd287cf4a30ed53d20700f714f821
 
 usage () {
     cat <<EOF
@@ -60,10 +59,9 @@ mkdir -p $DEPS
 
 if [ ! -d "$DEPS/smt-switch" ]; then
     cd $DEPS
-    git clone -b reset-reducer https://github.com/makaimann/smt-switch
+    git clone https://github.com/makaimann/smt-switch
     cd smt-switch
-    # temp pulling branch
-    # git checkout -f $SMT_SWITCH_VERSION
+    git checkout -f $SMT_SWITCH_VERSION
     ./contrib/setup-btor.sh
     if [ $cvc4_home = default ]; then
         ./contrib/setup-cvc4.sh
--- a/contrib/setup-smt-switch.sh
+++ b/contrib/setup-smt-switch.sh
@@ -3,8 +3,7 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 DEPS=$DIR/../deps
 
-# TEMP pull branch
-# SMT_SWITCH_VERSION=9230477a5b9cd287cf4a30ed53d20700f714f821
+SMT_SWITCH_VERSION=0c0b12a11b9a0d68d6d724389393120172ed50f3
 
 usage () {
     cat <<EOF
@@ -60,10 +59,9 @@ mkdir -p $DEPS
 
 if [ ! -d "$DEPS/smt-switch" ]; then
     cd $DEPS
-    git clone -b update-assuming-interface https://github.com/makaimann/smt-switch
+    git clone https://github.com/makaimann/smt-switch
     cd smt-switch
-    # TEMP pull branch
-    # git checkout -f $SMT_SWITCH_VERSION
+    git checkout -f $SMT_SWITCH_VERSION
     ./contrib/setup-btor.sh
     if [ $cvc4_home = default ]; then
         ./contrib/setup-cvc4.sh
--- a/contrib/setup-smt-switch.sh
+++ b/contrib/setup-smt-switch.sh
@@ -3,7 +3,8 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 DEPS=$DIR/../deps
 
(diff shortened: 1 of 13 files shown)


makaimann:
I'm starting jobs on the cluster to make sure this has the same performance increase seen in https://github.com/upscale-project/pono/pull/191

makaimann:
Marking `do not merge` until the runs finish and https://github.com/makaimann/smt-switch/pull/157 is merged and this is pointed back to the `smt-switch` `master` branch.

makaimann:
Screenshot from 2021-01-11 20-09-22

So this branch didn't do quite as well as the other one unfortunately. It got 94 vs 99 with the last full run of https://github.com/upscale-project/pono/pull/191. I still think this one is a bit cleaner. One potential difference is that this PR adds the implications for labels each time `rel_ind_check` is called. I'll try one more tweak on this branch to address that. Otherwise, we can decide if the extra solves are worth the less clean code and having to do all solving at context 1 when we use boolector with IC3IA.

engines/ic3base.cpp
```diff
@@ -101,14 +102,20 @@ static bool subsumes(const IC3Formula &a, const IC3Formula &b)
...
 
-IC3Base::IC3Base(const Property & p, const TransitionSystem & ts,
-                 const SmtSolver & s, PonoOptions opt)
+IC3Base::IC3Base(const Property & p,
+                 const TransitionSystem & ts,
+                 const SmtSolver & s,
+                 PonoOptions opt)
     : super(p, ts, s, opt),
-      reducer_(create_solver(s->get_solver_enum())),
+      // NOTE: this is a hack
```
ahmed-irfan, line 110: why is it a hack?

makaimann: Oops, my bad! That was an old comment from before I created the `create_reducer_for` method and was instead doing something hacky.

ahmed-irfan:
why https://github.com/upscale-project/pono/pull/191 is needed in this PR?

engines/ic3base.cpp
```diff
@@ -465,7 +477,12 @@ bool IC3Base::rel_ind_check(size_t i,
     for (const auto & cc : c.children) {
       ccnext = ts_.next(cc);
       lbl = label(ccnext);
-      solver_->assert_formula(solver_->make_term(Implies, lbl, ccnext));
+      if (lbl != ccnext && !is_global_label(lbl)) {
+        // only need to add assertion if the label is not the same as ccnext
+        // could be the same if ccnext is already a literal
+        // and is not alreayd in a global assumption
```
ahmed-irfan, line 483:
```suggestion
        // and is not already in a global assumption
```

engines/ic3ia.cpp
```diff
@@ -186,7 +185,9 @@ void IC3IA::abstract()
   // add predicates automatically added by ia_
   // to our predset_
   // needed to prevent adding duplicate predicates later
-  predset_.insert(bool_symbols.begin(), bool_symbols.end());
+  for (const auto & sym : bool_symbols) {
+    add_predicate(sym);
```
ahmed-irfan, line 189: why we are adding the boolean symbols? previously we were just adding them in predset_

ahmed-irfan:
left some comments

makaimann:
Screenshot from 2021-01-12 10-24-46

Here's the latest job. Looks better!

engines/ic3ia.cpp
makaimann: Right, good question. So I was worried about the tight integration with `ceg-prophecy-arrays`. With this change, the refinement step needs to not only add predicates back in but also add boolean symbols to both `predset_` and `predlbls_` or else there will be issues in `IC3IA`. I thought that was annoying and complicated so I just wrapped it all into `add_predicate` which checks if it's a symbol (or a literal more specifically) and does the right thing if so.

makaimann:
> why #191 is needed in this PR?

So this is an an alternative to #191 that seemed a bit cleaner. It doesn't actually use predicate variables in the frames, which makes it a bit easier to use `boolector` as well.

engines/ic3ia.cpp
ahmed-irfan: I see. Then, are you doing this change in Ceg-proph (in the refine-subprover-ts)?

ahmed-irfan:
> Screenshot from 2021-01-12 10-24-46
> 
> Here's the latest job. Looks better!

This looks great

contrib/setup-smt-switch.sh
```diff
@@ -59,9 +60,10 @@ mkdir -p $DEPS
 
 if [ ! -d "$DEPS/smt-switch" ]; then
     cd $DEPS
-    git clone https://github.com/makaimann/smt-switch
+    git clone -b update-assuming-interface https://github.com/makaimann/smt-switch
```
ahmed-irfan, line 63: reminder: change this

ahmed-irfan (approved):
LGTM

makaimann:
FYI had to make a change to fix an assertion failure when using `boolector`. Change is here 6d0f3364c8e6ab8c95111eefb8c2a58e5b2952fd. I tried to explain briefly in the comments but I'm happy to explain in more detail offline if needed.

makaimann:
> > > why #191 is needed in this PR?
> 
> > So this is an an alternative to #191 that seemed a bit cleaner. It doesn't actually use predicate variables in the frames, which makes it a bit easier to use `boolector` as well.
> 
> Oh, I meant to say that why [makaimann/smt-switch#157](https://github.com/makaimann/smt-switch/pull/157) is needed?

Gotcha, that's needed because `boolector` will complain about the labels. If we add the assumption `lbl <-> form` at the top-level, then it will think that `lbl` is not a boolean literal since it will treat it as equivalent to `form`.

engines/ic3ia.cpp
makaimann: Oops good point. Fixed with 4ccf3cde6e75b562306cb712d30a5743425862f7.

I cleared `predlbls_` in `refine-subprover-ts` in a previous commit.

PR #193, opened 2021-01-12
Add an option to assume the property in the pre-state

makaimann:


makaimann:
Screenshot from 2021-01-12 10-29-37

Here are the results. I ran it on top of a branch, but since this change is pretty much orthogonal to any backend algorithm, I think that's fine.

options/options.h
```diff
@@ -144,6 +146,7 @@ class PonoOptions
   static const bool default_cegp_axiom_red_ = true;
   static const std::string default_profiling_log_filename_;
   static const bool default_mod_init_prop_ = false;
+  static const bool default_assume_prop_ = false;
```
ahmed-irfan, line 149: how about turning it on by default.

ahmed-irfan (approved):
LGTM. 

It seems that this helps. Thanks for doing this.

Minor: what do you think about turn it on by default.

options/options.h
makaimann: Yeah I'm okay with that! Do you think we could run into any unsoundness issues by accident? Or as long as it's safety checking only we should always be good?

ahmed-irfan: I think that we are good for invariant properties and that's what we support

makaimann:
Oops, made a dumb mistake and didn't actually enable the option. That run was the same and difference must have just been noise.

makaimann:
Rebased on https://github.com/upscale-project/pono/pull/192 and force-pushed for evaluation. Currently running on cluster.

makaimann:
Screenshot from 2021-01-13 11-06-44

Unfortunately there's an unsoundness in this change. I'm not sure what's causing it yet, I'll look into it soon. Let me know if you have any thoughts on what it could be.

makaimann:
Nevermind, figured it out. The `assume_prop` pass was put before the prop monitor pass -- oops. I intentionally put it after `mod_init_prop` but forgot that there was a separate property modifying pass. Starting new jobs on the cluster now.

makaimann:
All right, that fix worked. Looks like there's a minor performance improvement:

Screenshot from 2021-01-13 15-09-18

ahmed-irfan:
thanks.

I think we should merge it even with this small improvement because it will definitely help ceg-proph.
What do you think?

makaimann:
Yeah I agree, I'll merge now.

PR #194, opened 2021-01-13
Strengthen prop in IC3 variants

makaimann:
Based on @zhanghongce's suggestion to use the original IC3 approach of finding bad cubes by checking with a transition while assume the property in the pre-state. Essentially this starts every fresh frame assuming the property, although the implementation doesn't do that because the property might not be representable by a single `IC3Formula` (e.g. the property might not be a clause).

This includes a few other edits. Most notably:
* it changes the `predecessor_generalization` interface to take a term instead of an `IC3Formula` for the term that we failed to block. This is because we might want to generalize the predecessor that can reach bad in one step. In that case, bad is the thing we failed to block but it might not be representable as an `IC3Formula` (e.g. usually a cube / clause).
* it gets rid of `cex_pg_` for recovering a counterexample trace and instead just reconstructs the trace immediately. I think this is better because now we have to add `bad_` to the end of the trace, and the function will take care of that for us

Overall I think this is better in terms of code quality in addition to performance. In particular I like that we now have the invariant that `reached_k_ == frontier_idx()`

### DIFF
--- a/engines/ic3base.cpp
+++ b/engines/ic3base.cpp
@@ -348,7 +348,7 @@ bool IC3Base::intersects_bad(IC3Formula & out)
 {
   push_solver_context();
   // assert the last frame (conjunction over clauses)
-  assert_frame_labels(reached_k_ + 1);
+  assert_frame_labels(frontier_idx());
   // see if it intersects with bad in next states
   solver_->assert_formula(ts_.next(bad_));
   // don't need transition relation for this check
@@ -398,14 +398,14 @@ ProverResult IC3Base::step(int i)
     return ProverResult::UNKNOWN;
   }
 
-  if (reached_k_ < 0) {
-    return step_0();
+  if (reached_k_ < 1) {
+    return step_01();
   }
 
   // reached_k_ is the number of transitions that have been checked
   // at this point there are reached_k_ + 1 frames that don't
   // intersect bad, and reached_k_ + 2 frames overall
-  assert(reached_k_ + 2 == frames_.size());
+  assert(reached_k_ == frontier_idx());
   logger.log(1, "Blocking phase at frame {}", i);
   if (!block_all()) {
     // counter-example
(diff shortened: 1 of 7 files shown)


makaimann:
Keeping this as a draft PR for now because I'm still running experiments on the cluster and the code needs to be cleaned up more before being merged.

makaimann:
Screenshot from 2021-01-13 11-11-54

Results look good. I just need to clean up the code a bit and then I'll make this a full PR.

makaimann:
@zhanghongce FYI for the interface changes and also let me know if you see any issues in this PR.

Additionally, I'm currently re-running jobs after the latest refactor to make sure I didn't introduce any issues.

zhanghongce:
Thanks, Makai! Will adapt SyGuS PDR accordingly.

engines/ic3base.cpp
```diff
@@ -164,6 +157,10 @@ void IC3Base::initialize()
   solver_->assert_formula(
       solver_->make_term(Implies, trans_label_, ts_.trans()));
 
+  // assume property in pre-state
+  Term prop = smart_not(bad_);
+  solver_->assert_formula(solver_->make_term(Implies, trans_label_, prop));
```
ahmed-irfan, line 162: isn't it already done in the other PR #193

makaimann: Yeah I thought about that. But wouldn't it break things if we turned off that option? Also, if someone runs it through the API the pass from #193 won't run automatically. I figured it wouldn't be too bad to assume it twice to avoid those issues.

ahmed-irfan: Moreover, I think we need to do the same in `reset-solver`

makaimann: Yeah that's here: https://github.com/upscale-project/pono/blob/af84199796cb4f492ce8a9642a3b661765811143/engines/ic3base.cpp#L952. I was thinking about doing it in a separate PR, but it would probably be good to put all the label assuming code in one function since it's done in two places.

ahmed-irfan: > Yeah I thought about that. But wouldn't it break things if we turned off that option? Also, if someone runs it through the API the pass from #193 won't run automatically. I figured it wouldn't be too bad to assume it twice to avoid those issues. What do you think?

that is a good point. Let's keep it the way it is now

ahmed-irfan (approved):

zhanghongce:
Hi Makai, I now have a question regarding this...

```
  Term prop = smart_not(bad_);
  solver_->assert_formula(solver_->make_term(Implies, trans_label_, prop));
```

This means when you use the labels, when asserting transition function, the property is also asserted (this in some sense makes property as an implicit lemma/constraint on that frame right?)

but on the hand, for the `get_frame_term` function, property is not factored in. So for those engines that also use this function like mbic3.cpp line 251, wouldn't you get a mismatch?

PR #196, opened 2021-01-14
CEGAR loop to abstract values

makaimann:
This PR adds initial support for a CEGAR loop that abstracts values. There are several limitations, but those can be removed with future PRs. In particular,

* it currently **only** supports `CegProphecyArrays<IC3IA>` as the backend prover, because that is the main motivation for this
* it currently abstracts **all** values. A future PR could allow a cutoff size

A few things I needed to change:
* since we are nesting CEGAR loops, we needed to specify which version of `cegar_abstract/cegar_refine` was running
* similarly, it was difficult to abstract `bad_` as-is because that happened before `initialize`. Instead, I just set `bad_` in the constructor initializer list

I ran into some unrelated issues here which are out-of-scope for this PR, but I left a TODO. The issue was with transferring the engine's view of an invariant to the original TS. We need to cache all the symbols, but it was not considering UFs. This came up because there are UFs in the invariant from the `CegProphecyArrays` abstraction.

### DIFF
--- a/engines/cegar_values.cpp
+++ b/engines/cegar_values.cpp
@@ -70,8 +70,8 @@ class ValueAbstractor : public smt::IdentityWalker
         // create a frozen variable
         Term frozen_var =
             ts_.make_statevar("__abs_" + term->to_string(), term->get_sort());
-        // make sure it's frozen
-        ts_.assign_next(frozen_var, frozen_var);
+        // will be made frozen later
+        // since the transition system will be modified after this
         save_in_cache(term, frozen_var);
         abstracted_values_[frozen_var] = term;
         return Walker_Continue;
@@ -155,8 +155,8 @@ void CegarValues<Prover_T>::initialize()
 
   // specify which cegar_abstract in case
   // we're inheriting from another cegar algorithm
-  // calls super::initialize within cegar_abstract
   CegarValues::cegar_abstract();
+  super::initialize();
 
   // update local version of ts over fresh solver
   cegval_ts_ = TransitionSystem(prover_ts_, to_cegval_solver_);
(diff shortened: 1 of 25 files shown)


engines/cegar_values.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+    for (auto con : conc_ts_.constraints()) {
+      // NOTE: there should be better infrastructure for re-adding constraints
+      // currently have to avoid re-adding the next version
+      if (conc_ts_.no_next(con)) {
+        prover_ts_.add_constraint(va.visit(con));
+      }
+    }
+  } else {
+    prover_ts_ = conc_ts_;
```
ahmed-irfan, line 226: prover_ts_ is pointing to conc_ts_. Is it intentional?

ahmed-irfan: It is not clear to me, what is happening here?

engines/cegar_values.cpp
```diff
@@ -0,0 +1,309 @@
...
+      }
+    }
+    return Walker_Continue;
+  }
+
+  TransitionSystem & ts_;
+  UnorderedTermMap & abstracted_values_;
+};
+
+TransitionSystem create_fresh_ts(bool functional, const SmtSolver & solver)
```
ahmed-irfan, line 99: how about putting this method in the ts.h as a static function?

pono.cpp
```diff
@@ -64,7 +61,9 @@ ProverResult check_prop(PonoOptions pono_options,
   Engine eng = pono_options.engine_;
 
   std::shared_ptr<Prover> prover;
-  if (pono_options.ceg_prophecy_arrays_) {
+  if (pono_options.cegp_abs_vals_) {
```
ahmed-irfan, line 64: we should check if the right options are set.
e.g. ceg-proph is set to true.

engines/cegar_values.cpp
makaimann: Ahh I'll leave a comment, it's a copy constructor so I'm just starting the `prover_ts_` with all the same behavior as `conc_ts_` before abstracting it.

makaimann: It's an easy way to avoid manually copying over all the state variables etc... (although I had to do that in the functional case anyway)

ahmed-irfan: I see. So, va will have the old ts reference. Right?

engines/cegar_values.cpp
makaimann: Great idea! It didn't work nicely to put in `ts.h` because it doesn't know about `Relational` or `Functional` `TransitionSystem`, but I created a new utility file in 6c7b9bc066942633f5447df8c9c8ce10ee1ce7da.

pono.cpp
makaimann: I check in the `make_cegar_values_prover` function. Do you think we should check here also?

ahmed-irfan: then it is fine.

engines/cegar_values.cpp
```diff
@@ -0,0 +1,301 @@
...
+    for (size_t i = 0; i < assumps.size(); ++i) {
+      if (core.find(assumps[i]) != core.end()) {
+        logger.log(2, "CegarValues adding refinement axiom {}", equalities[i]);
+        // need to refine both systems
+        cegval_ts_.add_constraint(equalities[i]);
+        Term transferred_eq =
+            from_cegval_solver_.transfer_term(equalities[i], BOOL);
+        // TODO this should be more modular
+        //      can't assume super::abs_ts_ is the right one to constrain
+        super::abs_ts_.add_constraint(transferred_eq);
```
ahmed-irfan, line 289: this is true for ic3ia and ceg-proph, but not for other engines.

We should specialize the refinement of the subprover ts, similar to what is done in the ceg-proph code.

makaimann: Good idea! Done in 182ddd9bb267fc0dac4027b638d0a593af15238a

makaimann:
Thanks for the review! I think I've addressed all the feedback so far. I'd also like to add a cutoff value for when to abstract the value, but since this is already a large PR I'll do that in a separate PR.

ahmed-irfan (approved):
LGTM.

makaimann:
Thanks for the reviews!

PR #197, opened 2021-01-14
Temporary configuration fix for CegProphecyArrays

makaimann:
As discussed on Zoom, this PR creates a temporary (hack) fix for the `--ceg-prophecy-arrays` option. The best configuration for `IC3IA` with `mathsat` as the underlying solver uses the `bool_model_generation` option, but unfortunately that breaks `CegProphecyArrays`. The long term fix is to use a fresh solver in `CegProphecyArrays` without that configuration but since that is fairly nontrivial, we are leaving that fix for later. For now, just disable that option if we're running `CegProphecyArrays`.

### DIFF
--- a/pono.cpp
+++ b/pono.cpp
@@ -166,8 +166,15 @@ int main(int argc, char ** argv)
   try {
     // no logging by default
     // could create an option for logging solvers in the future
-    SmtSolver s = create_solver_for(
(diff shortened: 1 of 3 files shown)


ahmed-irfan (approved):
LGTM

PR #201, opened 2021-01-15
Fix push/pop mismatch in array_axioms_refiner

makaimann:
I believe this never showed up before because we never used to continue execution after finding a counterexample. But now that we can have nested CEGAR loops, this is important.

Btw, we should probably discuss a better way of keeping track of solver contexts. I'm considering adding this to either smt-switch or having a wrapper solver that keeps track of extra info like that.


refiners/array_axiom_enumerator.cpp
```diff
@@ -233,6 +233,7 @@ bool ArrayAxiomEnumerator::enumerate_axioms(const Term & abs_trace_formula,
 
     if (!found_lemmas) {
       // there appears to be a concrete counterexample
+      solver_->pop();
```
ahmed-irfan, line 236: we may want to store the counterexample. Where do we do that? If we pop now, we will not have access to the model values.

makaimann: That's a good point. I believe it's currently not handled at all (I don't think witnesses were handled before either). I'm not sure that's the right place anyway though since it's in the refiner. Maybe we should unroll the concrete transition system to get the trace when there's a counterexample? And if we want to save abstract model values to help guide it we could do that too, but there's currently no infrastructure for that. Btw, `witness` should probably be overridden in `CegProphecyArrays` so that it returns the right witness there.

ahmed-irfan (approved):
LGTM. 

We discussed offline and decided to deal with the counterexample generation in a separate PR.

PR #202, opened 2021-01-16
Add property in get_frame_term

makaimann:
@zhanghongce thanks for catching this

### DIFF
--- a/engines/ic3base.cpp
+++ b/engines/ic3base.cpp
@@ -802,6 +802,9 @@ Term IC3Base::get_frame_term(size_t i) const
       res = solver_->make_term(And, res, u.term);
     }
   }
+
+  // the property is implicitly part of the frame
+  res = solver_->make_term(And, res, smart_not(bad_));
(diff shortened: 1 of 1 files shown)


makaimann:
Wait, I jumped the gun. Let me check a few things first.

makaimann:
All right, I think it's good to go.

ahmed-irfan:
is it a fix or an enhancement? I think it is the latter

ahmed-irfan (approved):

makaimann:
So I think it's actually a fix -- see here: https://github.com/upscale-project/pono/pull/194#issuecomment-761337288

Basically anything that uses the frame directly needs to have the property in it. I know that the frames are supposed to imply the property, but because we're always assuming the property, it's actually part of the frame. I.e. the frame starts as `P` and we add more clauses onto it so that there's no transition to bad. It's technically possible that `get_frame_term` would return something which doesn't imply `P` before this fix.

PR #205, opened 2021-01-16
refactor array abstactor modifier

ahmed-irfan:
This PR move the array abstraction- and concretization-walkers inside the Array Abstraction class.


makaimann (approved):
LGTM, thanks!

PR #206, opened 2021-01-16
Minor: Don't re-add property to invariant frame

makaimann:
Adding the property to the frame to get the invariant is no longer needed after https://github.com/upscale-project/pono/pull/202.


ahmed-irfan (approved):

makaimann:
Thanks for the review!

PR #207, opened 2021-01-18
Fix in mod-init-prop

makaimann:
I got a couple incorrect results using `--mod-init-prop` on some SMV benchmarks. It turns out it was related to an (incorrect) optimization. I thought that if the property was already a propositional literal, I didn't need to make a new one. However, due to the initstate transformation, it's important to use a delayed property regardless.


makaimann:
Marking `do not merge` until the cluster runs finish.

makaimann:
Discussed separately that we would use a different fix. Closing this PR.

PR #208, opened 2021-01-18
Create option for cutoff value of CegarValues

makaimann:
The cutoff is the magnitude of a value that will be abstracted. Anything of lower magnitude will be kept precise.

### DIFF
--- a/options/options.cpp
+++ b/options/options.cpp
@@ -54,7 +54,8 @@ enum optionIndex
   PROFILING_LOG_FILENAME,
   MOD_INIT_PROP,
   NO_ASSUME_PROP,
-  CEGP_ABS_VALS
+  CEGP_ABS_VALS,
+  CEGP_ABS_VALS_CUTOFF
 };
 
 struct Arg : public option::Arg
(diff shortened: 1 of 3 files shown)


options/options.cpp
```diff
@@ -271,6 +272,13 @@ const option::Descriptor usage[] = {
     Arg::None,
     "  --cegp-abs-vals \tabstract values in ceg-prophecy-arrays (only "
     "supported for IC3IA)" },
+  { CEGP_ABS_VALS_CUTOFF,
+    0,
+    "",
+    "cegp-abs-vals-cutoff",
+    Arg::Numeric,
+    "  --cegp-abs-vals-cutoff \tcutoff value for what to abstract (must be "
```
ahmed-irfan, line 280: shouldn't we also print the default value? (which I think is 100)

ahmed-irfan (approved):
LGTM.

Only one minor comment

options/options.cpp
makaimann: Good point! Made change with c1f5d53522acd2ab1d8181860a26881ff50edc83.

PR #210, opened 2021-01-19
Disable witness generation/printing by default

makaimann:
This PR turns witness printing off by default. This seems like better default behavior as opposed to dumping a potentially very long trace to the terminal.


lonsing (approved):
Looks good to me!

PR #211, opened 2021-01-19
Use witness length when refining in CegProphecyArrays

makaimann:


ahmed-irfan (approved):

ahmed-irfan:
do we need reached_k_ in the ceg-proph anymore?

makaimann:
> do we need reached_k_ in the ceg-proph anymore?

That's a good point! We might not -- I'll try removing it.

makaimann:
So I think the only reason we need it for now is there needs to be an initial value if we use the trick where we refine before calling the subprover. I like keeping it that way for now because it seems to help. Although when we integrate more tightly with the engine, it probably won't be necessary anymore. What do you think? Should I change that now?

makaimann:
Screenshot from 2021-01-19 11-24-13

Previous best in a development branch was `67` so it looks like this helps.

PR #212, opened 2021-01-19
Relational pseudo init and property transformation

makaimann:


makaimann:
Marking `do not merge` until https://github.com/upscale-project/pono/pull/213 is merged. I decided that the refactoring should be in a separate PR.

modifiers/mod_ts_prop.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "smt-switch/utils.h"
+#include "utils/logger.h"
+#include "utils/term_analysis.h"
+
+using namespace smt;
+using namespace std;
+
+namespace pono {
+
+TransitionSystem pseudo_init_and_prop(TransitionSystem & ts, Term & prop)
```
ahmed-irfan, line 29: can ts be const?

ahmed-irfan: and also prop?

ahmed-irfan (approved):
Left minor comments. 
Overall LGTM

modifiers/mod_ts_prop.cpp
makaimann: Hey, thanks for the review! So `prop` can't because it's edited in-place. And `ts` can't because of this line: https://github.com/upscale-project/pono/blob/c4e299d7b9d10b3f4da9aeec0cf98df86d954f8b/modifiers/mod_ts_prop.cpp#L35

Basically, if it's already a relational TS, we don't copy it. Instead it's just a reference to the original TS. But if you think that's over-optimized, I can just copy it for both cases.

PR #214, opened 2021-01-19
Add option to promote input variables to state variables

makaimann:


makaimann:
Built this on https://github.com/upscale-project/pono/pull/213 so marking `do not merge` until that's merged.

makaimann:
FYI @zhanghongce, once this is merged you could just require that this pass is run before running `SyGuS-PDR` -- or even just call `promote_inputvar` on all the input variables internally.

ahmed-irfan (approved):
LGTM. 

Please check the minor comments

core/ts.cpp
```diff
@@ -385,6 +386,21 @@ void TransitionSystem::add_inputvar(const Term & v)
...
 
+Term TransitionSystem::promote_inputvar(Term iv)
+{
+  assert(iv);
+  size_t num_erased = inputvars_.erase(iv);
+  if (!num_erased) {
+    throw PonoException("Cannot promote non-inputvars: " + iv->to_string());
+  }
+
+  assert(iv);
```
ahmed-irfan, line 397: we have a similar assert above. Is this needed?

pono.cpp
```diff
@@ -46,18 +46,81 @@ using namespace pono;
...
+       bad-state property. Based on that information, rebuild the
+       transition relation of the transition system. */
+    StaticConeOfInfluence coi(ts, { prop }, pono_options.verbosity_);
+  }
+
+  if (pono_options.promote_inputvars_) {
+    // this is a bit tricky. because promote_inputvar
+    // modifies inputvars_, should copy the set first
+    UnorderedTermSet inputvars = ts.inputvars();
+    for (auto iv : inputvars) {
```
ahmed-irfan, line 100: can it be `auto const & iv`?

core/ts.cpp
makaimann: Good point addressed with be917378730b7130df2ccc9959c8037134a79fe7

pono.cpp
makaimann: Good point! Addressed with be917378730b7130df2ccc9959c8037134a79fe7.

PR #216, opened 2021-01-20
Declare ProofGoalQueue in header

makaimann:


makaimann:
FYI @zhanghongce this just exposes the `ProofGoalQueue` so that you can use it for may-block proof goals.

zhanghongce:
Many thanks! That's great!

PR #217, opened 2021-01-20
IC3Base: Add option to disable unsat core generalization

makaimann:
Note for future reference: the unsat core generalization is in `rel_ind_check` because for some variants the extra generalization can help, and it's not in a separate function because it's less confusing in my opinion to not pop the context of a solver in a different function (and the unsat core has to be obtained within the same context).


makaimann:
@zhanghongce would you mind reviewing this PR when you get a chance? Would this work for you? You can set the option in the `initialize` method of `SyGuS-PDR`

zhanghongce:
Yes, it should work. Thanks a lot!

zhanghongce (commented):
Thanks! Will help with PR #145

PR #218, opened 2021-01-20
Start new frames with property

makaimann:
This is opposed to keeping property in pre-state of trans.

Because the property is not necessarily a valid `IC3Formula`, it is instead associated with the frame label, and added explicitly in `get_frame_term`. This has the impact of consistently representing frames as containing the property. It also has the side effect of not doing the (unnecessary) relative inductive check on the property for propagation.

The main motivation for this is that it's cleaner and more intuitive to associate the property with the frame than the transition relation.

### DIFF
--- a/engines/ic3base.cpp
+++ b/engines/ic3base.cpp
@@ -144,10 +144,6 @@ void IC3Base::initialize()
   solver_->assert_formula(
       solver_->make_term(Implies, trans_label_, ts_.trans()));
 
-  // assume property in pre-state
-  Term prop = smart_not(bad_);
-  solver_->assert_formula(solver_->make_term(Implies, trans_label_, prop));
-
   bad_label_ = solver_->make_symbol("__bad_label", boolsort_);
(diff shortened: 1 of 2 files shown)


makaimann:
Marking `do not merge` while I do a test run on the cluster.

ahmed-irfan:
i like this solution because in the end prop would not appear in the invariant produced by ic3.

ahmed-irfan (approved):

makaimann:
Screenshot from 2021-01-21 08-32-03

Here are the results of the cluster run. I'm not worried about number solved because it was a debug build, but there were no discrepancies and the only error was a refinement failure. Looks good to me so I'll merge!

makaimann:
FY @zhanghongce , hopefully this helps with https://github.com/upscale-project/pono/pull/145. The frames will start with the property in them from the base class.

zhanghongce:
Many thanks!

PR #219, opened 2021-01-20
Turn off assume-prop in pre-state by default

makaimann:
This PR sets `assume_prop_` to false by default. It can now be enabled with `--assume-prop`.

### DIFF
--- a/modifiers/mod_ts_prop.cpp
+++ b/modifiers/mod_ts_prop.cpp
@@ -100,4 +100,12 @@ TransitionSystem pseudo_init_and_prop(TransitionSystem & ts, Term & prop)
   return rts;
 }
 
+void prop_in_trans(TransitionSystem & ts, const Term & prop)
+{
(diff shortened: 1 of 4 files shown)


makaimann:
Hey @ahmed-irfan , I think disabling this option by default is the best thing to do right now. See https://github.com/upscale-project/pono/pull/145 for discussion on how it might not work correctly with SyGuS PDR. For now, I'd rather just have it off and we can always enable it manually for techniques that benefit from it. In the future we can re-evaluate the best default value for this option.

ahmed-irfan (approved):

makaimann:
Thanks for the review!

makaimann:
FYI @zhanghongce , `assume_prop_` is off by default now.

zhanghongce:
> FYI @zhanghongce , `assume_prop_` is off by default now.

Thanks!

PR #221, opened 2021-01-21
Improve Refinement in cegar values

ahmed-irfan:
This PR changes how the refinement in cegar-values is done. Essentially, it utilizes the refine-ts method from ceg-prophecy.


engines/cegar_values.cpp
```diff
@@ -317,37 +317,37 @@ bool CegarValues<Prover_T>::cegar_refine()
...
 template <>
 void CegarValues<CegProphecyArrays<IC3IA>>::refine_subprover_ts(
-    const Term & constraint)
+    const UnorderedTermSet & axioms)
 {
-  Term transferred_constraint =
-      from_cegval_solver_.transfer_term(constraint, BOOL);
-  assert(super::abs_ts_.only_curr(transferred_constraint));
-  super::abs_ts_.add_constraint(transferred_constraint);
+  super::refine_ts(axioms);
```
makaimann, line 350: Will this only work for another CEGAR loop? E.g. what happens in the future if we allow running `CegarValues` on something other than `CegProphecyArrays<IC3IA>`?

makaimann (approved):
Looks good to me! I just left a comment to think about for the future.

engines/cegar_values.cpp
ahmed-irfan: It should also work for other configurations of cegprophecy, e.g. CegProphecyArrays<Interpolantmc>

makaimann: Got it, but we'll have to adapt it if we wanted to run it over `k-induction` or `IC3IA` directly, right?

makaimann: Which is fine btw, we already would have to adapt it even before this change.

ahmed-irfan: Right

PR #224, opened 2021-01-21
Fix for constraints after promoting inputvars

makaimann:
This PR fixes a soundness bug that can show up when using option `--promote-inputvars`. This is because after promoting input variables to state variables with no next state update (e.g. implicit inputs), we should re-evaluate all the constraints. A constraint that was previously only added over current because it had inputs in it, should now be added over next as well. It was correct before the pass, because input variables are only relevant for a transition, but once they're promoted that's no longer true.

Since it is so unsafe to promote input variables directly, this PR removes that functionality of a `TransitionSystem`. Instead there is now a pass that (safely) rebuilds the transition system without input variables.

Note: This is of course a recurring issue, see https://github.com/upscale-project/pono/issues/73. I think there could be a better solution long term. Basically, we should pick a representation and stick with it. It is a bit tough because there are some algorithms that do better with all state variables, and others that make more sense to have "true" input variables. In principle, I like both for different reasons. The implicit inputs are the simplest and this is typically the formalism in papers. On the other hand, it's not very intuitive since inputs should not be part of the state representation, and I think it's nice to show syntactically that they're only important for transitions.

Regardless, for now, I think this is a reasonable fix. We can return to this when we have more time.

### DIFF
--- a/core/ts.cpp
+++ b/core/ts.cpp
@@ -386,19 +386,6 @@ void TransitionSystem::add_inputvar(const Term & v)
   name_term(v->to_string(), v);
 }
 
-Term TransitionSystem::promote_inputvar(Term iv)
-{
-  assert(iv);
-  size_t num_erased = inputvars_.erase(iv);
-  if (!num_erased) {
-    throw PonoException("Cannot promote non-inputvars: " + iv->to_string());
-  }
-
-  Term nv = solver_->make_symbol(iv->to_string() + ".next", iv->get_sort());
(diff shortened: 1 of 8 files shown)


makaimann:
FYI @zhanghongce, here's the fix I mentioned separately.

zhanghongce (commented):
Got it. Thanks, Makai!

ahmed-irfan (approved):
LGTM

We should revisit this discussion in near future.

Issue #225, opened 2021-01-22
MBIC3 issue


zhanghongce:
Hi Makai, this is actually not related to anything I'm working on, maybe not an issue at all, but just to let you know currently MBIC3 seems to be abnormal: producing invariants that are checked to be not inductive.

You can try with the `anderson.2.prop1-func-interl.btor2` in the sample.

This is what I got on my end, by running: `pono -e mbic3 -v 3 --check-invar anderson.2.prop1-func-interl.btor2`

```
Propagation phase at frame 5
INVARCHECK: init |= inv...OK
INVARCHECK: inv & trans |= inv'...FAIL
INVARCHECK: inv |= prop...OK
Invariant Check FAILED
unknown
b0
```

makaimann:
Thanks Hongce, I really appreciate it! This should fix it: https://github.com/upscale-project/pono/pull/226. I'll leave this issue open until that PR is merged. I'll also start a cluster run for `mbic3` (which I haven't been testing lately...).

makaimann:
Fixed.

makaimann:
![Screenshot from 2021-01-22 10-33-15](https://user-images.githubusercontent.com/23203473/105530900-59bb4c80-5c9d-11eb-98eb-aced095c319d.png)

No discrepancies after fix.

PR #226, opened 2021-01-22
Use default inductive_generalization for option 0 in mbic3

makaimann:


makaimann:
Hey @ahmed-irfan , this seems to fix https://github.com/upscale-project/pono/issues/225. Although I'm going to look into it more because I don't fully understand what was wrong about the previous implementation. Changing the solver also changed the outcome, but that could just be luck. Let me know if you have any thoughts.

makaimann:
Btw, I think this is a nice change even aside from the bug fix because there's less code duplication.

ahmed-irfan (approved):
LGTM

deduplication is always appreciated :)

makaimann:
In case you're curious, the issue was here: https://github.com/upscale-project/pono/blob/c6086aa4ac448bb1b5b9858d8a2f203e40cd16c4/engines/mbic3.cpp#L174. The label is actually for `ts_.next(t)` but it's using `label(t)` instead. I think the bug only occurred with `boolector` because of the aggressive rewriting. It seems like in most cases it's fine to use `t` instead of `ts_.next(t)` to get the label here, because they're always dropped together. But due to some rewriting going on, this can actually end up giving the wrong result.

PR #227, opened 2021-01-23
Ic3 split bits

makaimann:
This PR implements a very simple extension of IC3 which just splits bit-vector variables into their individual bits for the cubes / clauses in IC3. It does not do any bit-blasting of the rest of the system, instead relying on the SMT solver for that.


engines/ic3bits.h
```diff
@@ -0,0 +1,51 @@
...
+** Top contributors (to current version):
+**   Makai Mann
+** This file is part of the pono project.
+** Copyright (c) 2019 by the authors listed in the file AUTHORS
+** in the top-level source directory) and their institutional affiliations.
+** All rights reserved.  See the file LICENSE in the top-level source
+** directory for licensing information.\endverbatim
+**
+** \brief Bit-level IC3 implementation that splits bitvector variables
+**        into the individual bits for bit-level cubes/clauses
```
ahmed-irfan, line 13: Maybe it is worth to comment here that the algorithm (frames and cubes) work with bits but the transition system is still at the BV level

ahmed-irfan (approved):
LGTM. Just a minor comment

engines/ic3bits.h
makaimann: Good idea! Added with d6cda362861b7f07abc9e8f6cf0a361bfe658501.

makaimann:
Thanks for the review!

PR #228, opened 2021-01-23
Maintain contract that witness length refers to number of transitions

makaimann:
Within Pono, we consider the length of a trace to be the number of transitions. This was incorrect in the `IC3Base` implementation of `witness_length` and this PR fixes that.

### DIFF
--- a/engines/ceg_prophecy_arrays.cpp
+++ b/engines/ceg_prophecy_arrays.cpp
@@ -95,7 +95,8 @@ ProverResult CegProphecyArrays<Prover_T>::prove()
 
       if (res == ProverResult::FALSE) {
         // use witness length
-        reached_k_ = prover->witness_length();
+        // reached_k_ is the last k without a counterexample trace
+        reached_k_ = prover->witness_length() - 1;
       }
 
     } else {
@@ -142,7 +143,8 @@ ProverResult CegProphecyArrays<Prover_T>::check_until(int k)
 
       if (res == ProverResult::FALSE) {
         // use witness length
-        reached_k_ = prover->witness_length();
+        // reached_k_ is the last k without a counterexample trace
+        reached_k_ = prover->witness_length() - 1;
       } else if (res == ProverResult::TRUE) {
         try {
           // set the invariant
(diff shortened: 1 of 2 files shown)


engines/ceg_prophecy_arrays.cpp
```diff
@@ -142,7 +142,7 @@ ProverResult CegProphecyArrays<Prover_T>::check_until(int k)
 
       if (res == ProverResult::FALSE) {
         // use witness length
-        reached_k_ = prover->witness_length() - 1;
+        reached_k_ = prover->witness_length();
```
ahmed-irfan, line 145: I am thinking that it should be `witness_length() -1`

ahmed-irfan: because reached_k_ means that there is no bug up to that bound

ahmed-irfan: maybe we have to make the change elsewhere

engines/ic3base.cpp
```diff
@@ -198,7 +198,7 @@ size_t IC3Base::witness_length() const
 {
   // expecting there to have been a witness computed
   assert(cex_.size());
-  return cex_.size();
+  return cex_.size() - 1;
```
ahmed-irfan, line 201: as discussed offline, this  change is correct

engines/ceg_prophecy_arrays.cpp
makaimann: Oh interesting. Right. So I was referring to the witness length default implementation when talking about `reached_k_`. I hadn’t thought about this one. That does make sense that this should also subtract 1 because `reached_k_` should always be the last bound finished. I’ll make that change and make sure that `reached_k_` is used correctly for the refinement step.

makaimann: All right, it already worked -- good point. Made with 693002c7c7d6bf7ac57cbf45d3417e7c718a30be.

ahmed-irfan (approved):

PR #229, opened 2021-01-24
Add polarity axioms in CegarValues

makaimann:


engines/cegar_values.cpp
```diff
@@ -105,13 +107,29 @@ class ValueAbstractor : public smt::IdentityWalker
           return Walker_Continue;
         }
 
+        fresh_solver_->push();
```
ahmed-irfan, line 110: this seems like a hack to check if a value is positive or negative. Right? If so, please add a comment to improve this in future.

engines/cegar_values.cpp
```diff
@@ -269,6 +288,11 @@ void CegarValues<Prover_T>::cegar_abstract()
   super::bad_ = va.visit(super::bad_);
   cegval_bad_ = to_cegval_solver_.transfer_term(super::bad_, BOOL);
 
+  // add polarity axioms from abstraction
+  for (const auto & ax : va.polarity_axioms()) {
+    prover_ts_.add_constraint(ax);
```
ahmed-irfan, line 293: is it possible to add this to trans only?

makaimann: Good idea. Added here 266b7d06c797951eba0cf3706ffc6f8fb430b57e. Does that look all right?

engines/cegar_values.cpp
makaimann: Yeah added with 1d86cbb257beb95ab3b712739dc17bc20773633d

ahmed-irfan (approved):

PR #230, opened 2021-01-25
Ic3ia reset solver

ahmed-irfan:
This moves the refinement code from ceg-prophic3 to ic3ia. This change allows to use reset-solver for the refinement in other cegar<ic3ia>


makaimann (approved):
This looks good to me, but I wonder if there's a significant performance impact? It means traversing the whole transition relation every time we call `reset_solver`. Should we do some performance runs to check?

ahmed-irfan:
> This looks good to me, but I wonder if there's a significant performance impact? It means traversing the whole transition relation every time we call `reset_solver`. Should we do some performance runs to check?

That’s a good point.

ahmed-irfan:
If it is an issue, we could also think about changing when to call reset solver

makaimann:
> If it is an issue, we could also think about changing when to call reset solver

Right, or we could have a "soft" reset and a "hard" reset. More complicated but could probably still be done in a clean way. The hard reset could call the soft reset.

ahmed-irfan:
btw: how about I put this change in a separate method "reabstract()" inside ic3ia?

makaimann:
That sounds good to me!

ahmed-irfan:
ok. I have made the change. 

The reset-solver method is untouched. Instead, this PR introduces a new method called "reabstract" in IC3IA.

engines/ic3ia.cpp
```diff
@@ -341,6 +340,45 @@ bool IC3IA::is_global_label(const Term & l) const
   return true;
 }
 
+void IC3IA::reabstract()
+{
+  // predicates from init and bad
+  UnorderedTermSet preds;
+  get_predicates(solver_, ts_.init(), preds, false, false, true);
```
makaimann, line 347: One thing to note here. It looks a little funny to me to get the predicates from init and bad before calling `ia_.do_abstraction()`. It doesn't actually matter because this abstraction doesn't change init or bad, but just to make it more intuitive we could consider moving that after the re-abstraction. I don't feel strongly either way though.

makaimann (approved):
LGTM! Just left one comment.

engines/ic3ia.cpp
ahmed-irfan: updated in e4d900f

PR #231, opened 2021-01-26
UF operator abstraction

ahmed-irfan:


engines/cegar_ops_uf.cpp
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  // specify which cegar_abstract in case
+  // we're inheriting from another cegar algorithm
+  CegarOpsUf::cegar_abstract();
+  super::initialize();
+
+  // update local version of ts over fresh solver
+  cegopsuf_ts_ = TransitionSystem(prover_ts_, to_cegopsuf_solver_);
+  // update cache
```
makaimann, line 107: Just FYI this doesn't handle UFs in the formula (because they're not an input or state variable). This is a problem elsewhere in the code that we need to address later so I don't think it needs to be fixed now.

makaimann (approved):
LGTM, thanks for working on this! Left an FYI comment but I don't think any changes are needed.

PR #238, opened 2021-01-28
Fixes in IC3SA

makaimann:
This PR adds a fallback procedure for functional refinement. There's an issue where functional refinement can result in a value due to simplification in the smt-solver. In that case, it will fall back to a simpler refinement approach. Additionally, there are cases (especially when using actual input variables that cannot have lemmas learned over them) where functional refinement can add the same lemma again due to simplification. Another branch (`ic3sa-fixes`) is experimenting with using a logging solver to avoid that.

It also makes sure to learn the lemma which is important for making progress. To do this it uses a relation view of the main transition system. It also splits the constraints into the conjunctive partition for a better unsat core.

This PR also fixes an assertion failure that would show up occasionally (more often with mathsat) where the predecessor was over-generalized. There were several iterations to fix this including a recursive justify coi implementation (I think the important part is traversing as a tree instead of a DAG because location of the node matters), and various other attempts. In the end, the only thing that worked was adding all variables appearing in constraints (and in the state update of variables from next constraints). These fixes may be overly conservative, but at least they work and can be improved upon in a separate PR. There is good evidence that this assertion failure no longer occurs (several runs with different random seeds don't turn it up). Unfortunately this is highly non-deterministic so I can't be fully sure.

There are still some expected errors when using mathsat as the underlying solver. There appears to be a rewriting bug so that when we create an equality that should be guaranteed to be true in the model (because they had the same value) it ends up being a different term (e.g. rewritten to one that evaluates to false). This bug was already reported and there's not much to do now. However, this seems okay because boolector is better for this algorithm anyway.

### DIFF
--- a/engines/ic3sa.cpp
+++ b/engines/ic3sa.cpp
@@ -235,9 +235,9 @@ RefineResult IC3SA::refine()
 
   push_solver_context();
 
-  // TODO also use conjunctive partitions to split up constraints
-  // as much as possible
-
+  // due to simplifications can end up with the same terms
+  // for the constraints, avoid duplicating labels by keeping
+  // track with a set
   UnorderedTermSet used_lbls;
   TermVec lbls, assumps;
   Term unrolled;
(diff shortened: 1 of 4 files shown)


makaimann:
Screenshot from 2021-01-27 21-40-34

Results showing no discrepancies, no errors in the btor backend, and the only errors in the mathsat backend are the expected one that we're pretty sure is a mathsat bug.

PR #239, opened 2021-01-29
Adjust error vs unknown result

makaimann:
This PR makes it so that refinement failure in an IC3 variant returns `unknown` rather than throwing an exception. It also makes it so that a failed invariant check throws an exception instead of returning `unknown`. This is for the purposes of testing, where it makes sense to ignore refinement failures (which can happen) but not failed invariant checks which indicates a bug.


ahmed-irfan (approved):

makaimann:
Thanks for the review!

PR #240, opened 2021-01-29
Various improvements to IC3SA

makaimann:
Summary of changes:
* because of https://github.com/upscale-project/pono/pull/242, this removes the hack of adding all variables in constraints to the projection. Instead it just calls justify_coi on the constraints during predecessor generalization
* adds an option to not use functional refinement (sometimes value does better)
* adds option to use sequence interpolants to discover new terms
* adds option to adjust the amount of initial terms in the abstraction (not sure if this is useful or not, but it's there for experimentation. The default is all terms)
* adds an option to use a logging solver from smt-switch. This can help `IC3SA` in some cases because rewriting can remove useful terms.
* adds (somewhat hacky) support for `CegarOpsUf<IC3SA>`. This needs to be cleaned up in the future.

Unfortunately I wasn't able to add support for `--ceg-prophecy-arrays` with `IC3SA`. This is because `IC3SA` relies on a functional transition system and `CegProphecyArrays` sometimes adds relational axioms. This is an annoying technicality that should be surmountable, because `IC3SA` actually also adds relational axioms, but it needs to know that the transition system started out functional (so that functional unrolling is safe to do). I just don't think it's worth trying to hack in a fix for this right now.


makaimann:
FYI @ahmed-irfan, I'm working on some improvements to IC3SA in this branch. Once the `ic3sa` branch is merged, I'll make this a PR for the master branch. I'm going to add one or two more things to this branch, and then I should be done. I'll update the description once I've finished that.

makaimann:
Screen Shot 2021-01-29 at 11 33 39 AM

Here are some cluster results. Of course it's not always good to compare debug runs, but it's still something.

engines/cegar_ops_uf.cpp
```diff
@@ -39,15 +37,16 @@ CegarOpsUf<Prover_T>::CegarOpsUf(const Property & p,
...
-    to_cegopsuf_solver_(cegopsuf_solver_),
-    from_cegopsuf_solver_(super::prover_interface_ts().solver()),
-    cegopsuf_ts_(cegopsuf_solver_),
-    cegopsuf_un_(cegopsuf_ts_)
+    : super(p, create_fresh_ts(ts.is_functional(), solver), solver, opt),
+      conc_ts_(ts, super::to_prover_solver_),
+      prover_ts_(super::prover_interface_ts()),
+      oa_(conc_ts_, prover_ts_),
+      cegopsuf_solver_(
+          create_solver(solver->get_solver_enum(), opt.logging_smt_solver_)),
```
ahmed-irfan, line 45: is this going to create a logging solver if given in the ops

makaimann: Yeah that's right. Not the cleanest solution, but it's the best way (currently) to run this pass with boolector. It's only if enabled through the opts.

ahmed-irfan (approved):
LGTM

makaimann:
Thanks for the review! Much appreciated. I'm going to merge it now. 

Just fyi, I pushed one more small commit https://github.com/upscale-project/pono/pull/240/commits/305aee6f6ef5a4b61bd9f5202a4c8f6f56226453. It addresses the `reset_solver` issue where we can't call `reabstract` if the solver doesn't support `reset_assertions`. That change makes it so that using a logging solver with boolector also turns on the option to allow reset assertions.

Also, I realized that `CegarOpsUf` doesn't work for `IC3IA` and `boolector`, but I'm not going to worry about that for now. That takes slightly more work and `IC3IA` works best with `mathsat` anyway. Let me know if you have any concerns!

makaimann:
FYI again, I decided to undo that last commit anyway. Since `CegarOpsUf` and `boolector` is only supported with `IC3SA`, we don't need that option anyway. We can address this better in the future.

PR #243, opened 2021-02-02
SyGuS-PDR Bug Fix

zhanghongce:
Changes

* remove customized `reaches_bad` which uses a problematic block goal cache
* add manual limit to some SyGuS structures to avoid blow-up

### DIFF
--- a/engines/syguspdr.cpp
+++ b/engines/syguspdr.cpp
@@ -450,7 +450,6 @@ IC3Formula SygusPdr::inductive_generalization(
               base,
               pred_collector_.GetAllPredNext()
             );
-      lemma2cube_.emplace(ret.term, post_model);
 
       return ret;
     }
@@ -691,7 +690,7 @@ std::pair<IC3Formula, syntax_analysis::IC3FormulaModel *>
   }
 
   syntax_analysis::IC3FormulaModel * partial_model = 
-    new syntax_analysis::IC3FormulaModel(std::move(cube_partial), conj_partial, false);
+    new syntax_analysis::IC3FormulaModel(std::move(cube_partial), conj_partial);
 
   assert(partial_model);
   model2cube_.emplace(conj_partial, partial_model);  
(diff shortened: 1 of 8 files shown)


makaimann (approved):
Looks good to me, thanks for the quick fix! Much appreciated.

zhanghongce:
Hi Makai, I'm still running some tests to see if there are other issues...

makaimann:
Travis has been slow lately so I'm just going to merge. I ran the tests locally and it seems to be a fairly straightforward change anyway. If it fails on other OSes for some reason, I can fix it.

zhanghongce:
Sorry, this is not the right fix. I will need more time to figure it out...

makaimann:
No worries, let me know how I can help.

zhanghongce:
Hi Makai,

I think I figured out the problem. This was because of incorrect constraint handling (the total effect of this error was not using the prime version of the constraint). I'm a bit surprised it only caused one discrepancy.

Also, to be safe, I think it is better to also keep the first two commits which un-do an incorrect optimization.

I'm now running more tests and see if there is any other issue caused by this change.

Thanks
Hongce

makaimann:
Thank you for looking into this! So just to check my understanding, the issue was the `nxt_state_updates_` was not yet populated, so using `next_curr_replace` on the next version of the constraints did nothing?

zhanghongce:
> Thank you for looking into this! So just to check my understanding, the issue was that `nxt_state_updates_` was not yet populated, so using `next_curr_replace` on the next version of the constraints did nothing?

Yes, I think so.

zhanghongce:
Another thing, Makai, I'm thinking of exposing a configuration to the commandline, so if there is more available resources, we can try and see if that will make it solve more instances. Is it okay?

makaimann:
> Another thing, Makai, I'm thinking of exposing a configuration to the commandline, so if there is more available resources, we can try and see if that will make it solve more instances. Is it okay?

Yes, that's perfect! We still have time. I have one thought to bring up with you by the way. 

I noticed that the children you pass to `IC3Formula` is just the single, full term: https://github.com/upscale-project/pono/blob/c175a302857ff00229a0919d5cc8fc3f78d04a26/engines/syguspdr.cpp#L316 
This is fine of course, and since you implement your own `predecessor_generalization` and `inductive_generalization`, it shouldn't matter. But, I was wondering if you might benefit from the subsumption check when constraining frames? I think that's the only place where it would matter, but I'm not sure if you would expect to benefit from that. Just thought I'd bring it up.

zhanghongce:
Thanks for bringing this up. Previously I break the second argument. And then, I changed it to the current version in the hope of reducing some overhead in `rel_ind_check` (because I'm not using the UNSAT core reduction in rel_ind_check and things like  labels for the children are not in use). However, I found that it is not better nor worse. I can change it back if you think it would be better.

makaimann:
>  However, I found that it is not better nor worse. I can change it back if you think it would be better.

Gotcha, I don't mind either way -- whichever you prefer!

makaimann:
@zhanghongce are you still running tests, or should I merge this?

zhanghongce:
> @zhanghongce are you still running tests, or should I merge this?

It is now finished. Thanks!

PR #244, opened 2021-02-03
SyGuS-PDR option

zhanghongce:
This exposes `--sygus-term-mode` option. 
Range 0-4, default: 4 auto. But I think sometimes "auto" does not give the best performance.


makaimann (approved):
Looks great, thanks for the option!

PR #245, opened 2021-02-03
IC3SA: traverse formula iteratively as DAG in justify_coi

makaimann:


engines/ic3sa.cpp
```diff
@@ -816,43 +816,58 @@ UnorderedTermSet IC3SA::add_to_term_abstraction(const Term & term)
...
+          to_visit.push_back(state_updates.at(curr_c));
+        }
       }
-    }
 
-    UnorderedTermSet free_vars;
-    get_free_symbolic_consts(c, free_vars);
-    for (const auto & fv : free_vars) {
-      projection.insert(fv);
+      UnorderedTermSet free_vars;
```
ahmed-irfan, line 866: minor: declare this outside of the loop and clear it here

ahmed-irfan (approved):
LGTM

only one minor comment

engines/ic3sa.cpp
makaimann: Good point -- addressed with a1222d000e801ac16210a389e4c769ed5acda662.

ahmed-irfan: even better :)

PR #248, opened 2021-03-02
Add test for modifiers in python interface

makaimann:


makaimann:
Had this test locally on my laptop so I thought it would be good to include it.

ahmed-irfan (approved):

PR #253, opened 2021-04-15
Allow quoted symbol names in SMV

makaimann:
This PR would allow symbol names in quotes that have characters which are not normally allowed in identifiers, such as `[` or `!`.


lonsing (approved):
Looks good to me!

PR #258, opened 2021-05-03
Seed grammar with maxterms + other fixes

makaimann:


makaimann:
Marking `do not merge` until maxterms are exposed as optional to command line.

PR #259, opened 2021-05-03
Try cheap-ish check of known predicates before going to sygus

makaimann:


makaimann:
Helps a noticeable amount but we're still behind. But using these predicate candidates and the all-sorts option gives us the most overall solves:

Screen Shot 2021-05-04 at 1 39 10 PM

PR #260, opened 2021-05-03
Don't break early if searching subterms in get_predicates

makaimann:


makaimann:
No disagreements and difference in number solved likely just luck. This change shouldn't actually affect ic3ia anyway because `search_subterms` is off.

Screen Shot 2021-05-04 at 1 35 21 PM

PR #264, opened 2021-05-20
Keep reference to TS in encoders at python level

makaimann:
This PR keeps a reference to the TS in the encoders at the Python level in the Python bindings. This is important, because otherwise the Python garbage collector doesn't know the relationship and might delete out-of-order which can lead to segmentation faults.


lonsing (approved):

makaimann:
Thanks for the review!

PR #265, opened 2021-05-21
Update smt switch version

makaimann:
This PR updates the smt-switch version used in Pono. It temporarily disables any tests that use two CVC4 instances, due to https://github.com/cvc5/cvc5/issues/5893. This limitation will be removed once that issue is resolved, and the updated version of (now) cvc5 is available through smt-switch.

### DIFF
--- a/tests/test_ic3.cpp
+++ b/tests/test_ic3.cpp
@@ -76,7 +76,12 @@ TEST_P(IC3UnitTests, SimpleSystemUnsafe)
   ASSERT_EQ(r, FALSE);
 }
 
-INSTANTIATE_TEST_SUITE_P(ParameterizedSolverIC3UnitTests,
-                         IC3UnitTests,
-                         testing::ValuesIn(available_solver_enums()));
+INSTANTIATE_TEST_SUITE_P(
+    ParameterizedSolverIC3UnitTests,
+    IC3UnitTests,
+    // TEMP excluding CVC4 because IC3 variants use two solvers
+    // and this causes strange behavior for CVC4 until its fixed
+    // (note: updated to cvc5 now)
+    // see https://github.com/cvc5/cvc5/issues/5893
+    testing::ValuesIn(available_solver_enums_except({ smt::CVC4 })));
 }  // namespace pono_tests
--- a/contrib/setup-smt-switch.sh
+++ b/contrib/setup-smt-switch.sh
@@ -3,7 +3,7 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 DEPS=$DIR/../deps
 
-SMT_SWITCH_VERSION=c3957b2e7fec8af4fab24df089230c88e523c0e6
+SMT_SWITCH_VERSION=a2af6ea75b994d666c9c175793a70c8945208ba5
 
 usage () {
     cat <<EOF
--- a/engines/ceg_prophecy_arrays.cpp
+++ b/engines/ceg_prophecy_arrays.cpp
@@ -412,7 +412,7 @@ void CegProphecyArrays<Prover_T>::reduce_consecutive_axioms(
   assert(res.is_unsat());
 
   UnorderedTermSet core;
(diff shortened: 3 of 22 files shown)


tests/test_ic3.cpp
```diff
@@ -76,7 +76,12 @@ TEST_P(IC3UnitTests, SimpleSystemUnsafe)
   ASSERT_EQ(r, FALSE);
 }
 
-INSTANTIATE_TEST_SUITE_P(ParameterizedSolverIC3UnitTests,
-                         IC3UnitTests,
-                         testing::ValuesIn(available_solver_enums()));
+INSTANTIATE_TEST_SUITE_P(
+    ParameterizedSolverIC3UnitTests,
+    IC3UnitTests,
+    // TEMP excluding CVC4 because IC3 variants use two solvers
```
lonsing, line 82: Is this an issue in the tests only or are we also creating two instances of CVC4 anywhere in Pono?

lonsing (approved):
Looks good to me! I left one comment regarding the issue with having two solver instances of CVC4/5.

However, I was not able to build with Python bindings but could not track down the problem.

This is the output of running `./contrib/setup-smt-switch.sh --python`:

```
....
-- Found Threads: TRUE  
CMake Warning at CMakeLists.txt:103 (find_package):
  By not providing "FindPython3.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Python3", but
  CMake did not find one.

  Could not find a package configuration file provided by "Python3" with any
  of the following names:

    Python3Config.cmake
    python3-config.cmake

  Add the installation prefix of "Python3" to CMAKE_PREFIX_PATH or set
  "Python3_DIR" to a directory containing one of the above files.  If
  "Python3" provides a separate development package or SDK, be sure it has
  been installed.


-- Found GMP: /usr/include/x86_64-linux-gnu  
-- Found GMP libs: /usr/lib/x86_64-linux-gnu/libgmp.so
-- Found GMPXX libs: /usr/lib/x86_64-linux-gnu/libgmpxx.so
-- Didn't find python library and include directory -- falling back
FATALCannot locate python library...
CMake Error at python/CMakeLists.txt:42 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.


CMake Error at python/CMakeLists.txt:43 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.


CMake Error at python/CMakeLists.txt:45 (list):
  list GET given empty list


CMake Error at python/CMakeLists.txt:46 (list):
  list GET given empty list


CMake Error at /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PythonInterp (missing: PYTHON_EXECUTABLE) (found version "")
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.10/Modules/FindPythonInterp.cmake:152 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  deps/scikit-build/skbuild/resources/cmake/FindPythonExtensions.cmake:247 (find_package)
  python/CMakeLists.txt:56 (include)


-- Configuring incomplete, errors occurred!
```

tests/test_ic3.cpp
makaimann: It could happen through the command line too, actually. I can add a check for that. If the user requests cvc4 with certain engines then we could end up creating two instances. Thanks for checking that!

makaimann:
Thanks for the review! Regarding that error message. It looks like this is a CMake version issue. There are two solutions: 1) increase the required CMake version or 2) change back to an old CMake module. I'm leaning towards the latter for now, because the cmake FindPython3 module we're using right now does not integrate very well with scikit-build yet anyway. We can always update it in the future (since FindPythonInterp is technically deprecated). These PRs would make the changes needed for 2:
https://github.com/makaimann/smt-switch/pull/244
https://github.com/upscale-project/pono/pull/267

PR #274, opened 2021-05-28
Throw exception if attempting to use cvc4 with a multi-solver engine

makaimann:
This PR would throw an exception if trying to use CVC4 with an engine that uses multiple instances of the solver. This is only temporary until the multi-solver issues in CVC4 are resolved.

### DIFF
--- a/options/options.h
+++ b/options/options.h
@@ -40,7 +40,7 @@ enum Engine
   IC3SA_ENGINE,
   SYGUS_PDR
   // NOTE: if adding an IC3 variant,
-  // make sure to update ic3_variants in smt/available_solvers.cpp
+  // make sure to update ic3_variants_set in options/options.cpp
   // used for setting solver options appropriately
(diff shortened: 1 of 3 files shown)


options/options.h
```diff
@@ -44,6 +44,15 @@ enum Engine
   // used for setting solver options appropriately
 };
 
+// NOTE keep this up to date so that setting
```
lonsing, line 47: Please update also the comment in lines 42-44 above to refer to `ic3_variants_set` in `./options/options.cpp`, which must be updated when adding IC3 variants.

lonsing (approved):
Looks good to me! 

I left one remark regarding an outdated comment.

options/options.h
makaimann: Good catch! Fixed in e847cde3a673d90b0387600cf95b5ccee27b35ac.

PR #283, opened 2021-07-02
Stream printing for ProverResult and engines

makaimann:


options/options.cpp
```diff
@@ -574,4 +574,61 @@ ProverResult PonoOptions::parse_and_set_options(std::vector<std::string> & opts,
...
+    case IC3SA_ENGINE: {
+      res = "ic3sa";
+      break;
+    }
+    case SYGUS_PDR: {
+      res = "sygus-pdr";
+      break;
+    }
+    default: {
+      throw PonoException("Unhandled engine: " + std::to_string(e));
```
lonsing, line 622: If `e` has the value `IC3_BOOL` then an exception will be thrown. Is that the intended handling of that case?

lonsing (approved):
Looks good to me.

I just left one comment.

options/options.cpp
makaimann: Nope, that was definitely a mistake. Good catch, thanks! Fixed with 681cc834fc07612970728cda14d5c68dee24b858

lonsing: Thanks! I was not sure whether that case can actually occur (at least, when setting options from the command line) because `to_engine(std::string s)` as used in option parsing does not include `IC3_BOOL`.

makaimann: Right, I haven't exposed `IC3_BOOL` to the command line yet, but I think it's still good to be able to print it (even just for failed unit tests that print which engine it was). Additionally, it probably makes sense to expose it, because a user could pass a boolean model through SMV.

PR #284, opened 2021-07-02
Fix non-consecutive axiom reduction in CEGP

makaimann:


makaimann:
Results are not actually better surprisingly.

Original results:
ref

Updated results (with more configurations):
cegp-reduce-fix

makaimann:
Looks better with old version of MathSAT.

cegp-msat-5 6 3

PR #285, opened 2021-07-02
Fix edge case in logger string formatting

makaimann:


lonsing (approved):
Looks good to me.

makaimann:
Thanks for your review!

PR #292, opened 2021-08-17
IC3IA: Continue even if predicate reduction fails

makaimann:
Ran into this running Pono with:
./pono -v 1 -e ic3ia --smt-solver msat --ceg-prophecy-arrays [chc-LIA-lin-arrays_247.smt2](https://github.com/chc-comp/chc-comp20-benchmarks/blob/master/LIA-lin-arrays/chc-LIA-lin-arrays_247.smt2.gz).

It does not seem like this should be possible, so that will take more investigation. There's nothing obviously wrong, because we've evaluated IC3IA on lots of other benchmarks without hitting this case.


makaimann:
Waiting to merge until after https://github.com/upscale-project/pono/pull/291.

PR #301, opened 2022-08-23
BMC options for improved bug hunting

lonsing:
This PR adds several options to modify the behavior of BMC. By default, the transition system is unrolled step by step to explore all possible bounds to find a counterexample, starting from zero. This leads to as many solver calls as there are bounds to be checked.

To reduce the number of solver calls and (optionally) skip over certain bounds, the options added in this PR allow to configure BMC to explore intervals of bounds. If a counterexample is found within a given interval, then either binary search or linear search is applied to the interval to find the precise bound of the counterexample.

Optionally, completeness of BMC to find counterexamples can be switched off to potentially speed up a BMC run by never checking certain bounds.

### DIFF
--- a/engines/bmc.cpp
+++ b/engines/bmc.cpp
@@ -112,7 +112,7 @@ bool Bmc::step(int i)
 
   solver_->push();
 
-  const int cex_guarantee = 1;
+  const int cex_guarantee = !options_.bmc_single_bad_state_;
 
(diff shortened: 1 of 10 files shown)


lonsing:
Merge this PR first to fix failing CI runs on MacOS: https://github.com/upscale-project/pono/pull/302

Issue #55, opened 2020-07-14
Look into infrequent errors in interpolation


makaimann:
Mentioned here: https://github.com/upscale-project/pono/pull/53

It's possible this is in MathSAT in which case there's not much we can do. But we should be sure that's the case.

lonsing:
This merged PR might be related: https://github.com/upscale-project/pono/pull/306

There were segfaults when running IC3IA engine caused by different versions of MathSAT used in Pono and smt-switch, which were fixed by updating to MathSAT 5.6.4 in Pono. However, the latest version is 5.6.8 (as of September 2022; `https://mathsat.fbk.eu/download.html`)

PR #306, opened 2022-09-15
Fix installation of IC3ia and update MathSAT

lonsing:


lonsing:
- The URL of IC3IA in `./contrib/setup-ic3ia.sh` was broken due to upstream update (previous version no longer available).
- We now download the new version: `https://es-static.fbk.eu/people/griggio/ic3ia/ic3ia-22.07.tar.gz`
- This change caused many segfaults when running IC3IA, so we also updated MathSAT to be the same version (5.6.4) as in smt-switch: https://github.com/stanford-centaur/smt-switch/blob/499621b009ca0e86a65339aed59157bfaa874776/ci-scripts/setup-msat.sh#L45-L52
- Updating MathSAT fixed the segfaults

This issue might be related to which versions of MathSAT are used in Pono and smt-switch: https://github.com/upscale-project/pono/issues/55

PR #325, opened 2024-06-18
Create virtual environment for Python packages in CI

CyanoKobalamyne:
Fixes CI failures due to [PEP 668](https://peps.python.org/pep-0668/).


CyanoKobalamyne:
CI is still failing, but at least it gets past the Python install step now.

PR #341, opened 2024-08-21
Exclude state variables without update functions in simple path check

CyanoKobalamyne:
If there are state variables that are, e.g., used in the property, but do not have an update function, the simple path check in k-induction  and BMC will never succeed. This is because we can always pick an arbitrary (new) value for this variable and thus construct a "new" state, even if, in reality, we are in a loop.

Such variables should in reality be inputs. However, such a transformation is not completely straightforward, as they could still have an initial value. For now, as a simple solution, we simply exclude them from the simple path checks.

### DIFF
--- a/core/ts.cpp
+++ b/core/ts.cpp
@@ -628,6 +628,17 @@ bool TransitionSystem::no_next(const Term & term) const
   return contains(term, UnorderedTermSetPtrVec{ &statevars_, &inputvars_ });
 }
 
+smt::UnorderedTermSet TransitionSystem::no_next_states() const
+{
+  smt::UnorderedTermSet no_next_states;
+  for (const auto & sv : statevars_) {
+    if (state_updates_.find(sv) == state_updates_.end()) {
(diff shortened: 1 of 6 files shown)


CyanoKobalamyne:
It took me a while to update everything correctly, but now it should be computed whenever the rest of the internal structures are updated.

Also, it seems that with this change k-induction (and BMC with simple path checking) can now prove the property in `InterpWinTests`.

barrettcw (approved):
LGTM

ahmed-irfan:
Cool feature!

One suggestion: I think this feature should be independent of whether Cone
of Influence (COI) is enabled or not. Currently, if I am not mistaken, it
is tied to rebuilding the transition system with COI.

Great job!!!

CyanoKobalamyne:
This should actually work regardless of whether COI analysis is enabled. But I also changed the COI rebuilding process so that it rebuilds the list of state variables with no next states, otherwise this would break when static COI analysis is enabled.

PR #343, opened 2024-08-29
Add option to create printing solver

CyanoKobalamyne:
This allows the user to set an option, `--printing-smt-solver`, that will result in a solver instance being constructed using the `create_printing_solver` function in smt-switch. This will dump all formulas sent to the solver by Pono in SMT-LIB format. This allows one to replay the queries in the standalone solvers and is useful for debugging.


CyanoKobalamyne:
Also fixed some related comments and header includes.

PR #371, opened 2025-03-18
Support SMV bit-vector resize operators

augustomafra:
SMV files generated from Yosys use several bit-vector resize operations that are not currently supported by pono.

After implementing them and testing on a simple 2-bit Verilog counter compiled with Yosys, I hit a soundness bug that seemed to be affecting RelationalTransitionSystems only (hence finding it only after compiling to SMV, as previous Btor2 models were creating FunctionalTransitionSystems).


augustomafra:
Hi, @CyanoKobalamyne ,

Could you please take a look at the pull request? It would be helpful to get feedback about my fix on kinduction.cpp on commit [c9bc82f](https://github.com/stanford-centaur/pono/pull/371/commits/c9bc82f2262796f037e29b44a7bf2395d2d90249). It does solve the soundness bug I found in my test case, but I'm not sure if I should dig deeper into the root cause of the trivial "assert false" constraint appearing in simple_path_constrain in first place.

Let me know if there is anything else I can help for the pull request, thanks!
Augusto

CyanoKobalamyne:
Hi! Thanks for the contribution. Sorry that we still have not gotten around to reviewing this. We are currently preparing for the Hardware Model-Checking Competition, which only features benchmarks in BTOR2 format, so improving SMV support will only be addressed after that.

In the meantime, could you split this into two pull requests, one containing only the SMV syntax-related changes, and the other one the proposed fix for the RTS bug? And please enable the CI workflows (actions) on your fork so that we can see the status checks. Thanks!

augustomafra:
Hi, thanks for the response! Sure, I did several other changes recently for my research project, so I still need to take some time to organize the pull requests with the new commits. I will update the CI for the upcoming changes.

PR #372, opened 2025-03-18
Add an option "--version" to print version and exit

Po-Chun-Chien:
This PR introduces a new option `--version`:

```
  --version  Print version and exit.
```

Example usage:

```bash
$ ./pono --version
v0.1.1-237-g35cadb4
```

The version is obtained from running `git describe --always --dirty`.
This feature is useful if one wants to determine the version of the code from which the binary is compiled.


CyanoKobalamyne:
Would you mind rebasing this? Thanks!

CyanoKobalamyne (approved):

PR #373, opened 2025-03-19
Add an option to dump Btor2 witness to file

Po-Chun-Chien:
Example usage:

```bash
./pono --dump-btor2-witness witness.cex path/to/file.btor2
```

This option implicitly turns on the `--witness` option.
If only `--witness` is specified but `--dump-btor2-witness` is not given,
then the witness will be printed to stdout (the same behavior as before).


utils/logger.h
```diff
@@ -138,6 +138,23 @@ class Log
...
+   * @param level the verbosity level to print this log (prints for any
+   * verbosity greater than this level)
+   * @param format the format string
+   * @param args comma separated list of inputs for the format string
+   */
+  template <typename... Args>
+  void log_to_stream(size_t level,
+                     std::ostream & output_stream,
+                     const std::string & format,
+                     const Args &... args) const
```
Po-Chun-Chien, line 151: We can also overload this function as `log`. I do not have a strong preference.

ahmed-irfan (commented):
LGTM

Can you fix the conflicts? I will do another pass after that.

Po-Chun-Chien:
> Can you fix the conflicts? I will do another pass after that.

I've merged `main` and fixed the conflicts.

ahmed-irfan (approved):

PR #375, opened 2025-03-20
Update IC3ia to version 23.05

Po-Chun-Chien:
The old version 22.07 of IC3ia is no longer available for download.

The new IC3ia 23.05 can be built with Pono locally on my laptop without a problem.

### DIFF
--- a/contrib/setup-ic3ia.sh
+++ b/contrib/setup-ic3ia.sh
@@ -25,7 +25,7 @@ mkdir -p $DEPS
 cd $DEPS
 
 msat_home="$DEPS/mathsat"
-ic3ia_version=ic3ia-22.07
+ic3ia_version=ic3ia-23.05
 
 while [ $# -gt 0 ]
 do


Po-Chun-Chien:
The pipeline is failing because mathsat cannot be downloaded.
The [download link](https://mathsat.fbk.eu/download.php?file=mathsat-5.6.4-linux-x86_64.tar.gz) for 5.6.4 does not seem to work at the moment.
(The [link](https://mathsat.fbk.eu/download.php?file=mathsat-5.6.11-linux-x86_64.tar.gz) does not work either.)
There is likely some issue with their sever.

CyanoKobalamyne:
Yeah, I saw, the other CI workflows are failing as well.

Po-Chun-Chien:
> Yeah, I saw, the other CI workflows are failing as well.

I will notify them and see if this can get fixed quickly.

CyanoKobalamyne:
We didn't update the MathSAT version because there were some regressions w.r.t. 5.6.4, but we can take a look again.

Po-Chun-Chien:
> We didn't update the MathSAT version because there were some regressions w.r.t. 5.6.4, but we can take a look again.

I see. I'll revert that then.
We can re-evaluate MathSAT versions later.

CyanoKobalamyne:
Thanks!

CyanoKobalamyne (approved):

PR #356, opened 2024-09-25
Adjust bound and witness when monitor state is added to TS

CyanoKobalamyne:
Fixes #320.

PR #378, opened 2025-03-26
Check if user-specified bound is valid

Po-Chun-Chien:
This PR adds a simple check for the user-specified bound (`--bound`),
ensuring it is less than `INTMAX`.

The upper limit is set to prevent potential integer overflow when a monitor is used,
as the bound is internally incremented by one (see #356).

### DIFF
--- a/options/options.cpp
+++ b/options/options.cpp
@@ -670,7 +670,7 @@ ProverResult PonoOptions::parse_and_set_options(int argc,
         case ENGINE: engine_ = to_engine(opt.arg); break;
         case BOUND:
           bound_ = atoi(opt.arg);
-          if (bound_ >= INT_MAX) {
+          if (bound_ == INT_MAX) {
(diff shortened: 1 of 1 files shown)


options/options.cpp
```diff
@@ -667,7 +668,13 @@ ProverResult PonoOptions::parse_and_set_options(int argc,
         case HELP:
           // not possible, because handled further above and exits the program
         case ENGINE: engine_ = to_engine(opt.arg); break;
-        case BOUND: bound_ = atoi(opt.arg); break;
+        case BOUND:
+          bound_ = atoi(opt.arg);
+          if (bound_ >= INT_MAX) {
```
CyanoKobalamyne, line 673:
```suggestion
          if (bound_ == INT_MAX) {
```
Maybe change it to `INT_MAX`, because anything above that is undefined behavior.

Issue #377, opened 2025-03-26
Wrong proofs caused by property monitor


Po-Chun-Chien:
Pono (at commit aa4b19d9) produces a wrong proof (unsat) on [bv_debug_16.btor2](https://gitlab.com/sosy-lab/research/data/word-level-hwmc-benchmarks/-/blob/fe579e59fc21687e4a0dcbdb4988178007919d48/bv/btor2/fuzzbtor2/bv_debug_16.btor2),
whereas both AVR and BtorMC report sat.

The issue occurs with the engines `ind`, `ic3ia`, `ic3bits`, and `interp` (but interestingly not with `ic3sa`).

```
./pono --engine ind bv_debug_16.btor2
unsat
b0
```

## Cause of the Unsoundness

After some investigation, I found out that the unsoundness is caused by the combined effect of property monitor and `constraint` in Btor2.
When the property signal (`bad` in Btor2) depends on an input,
Pono creates a monitor state and store the property value into it.
The next-state value of the monitor reflects the property evaluated at the current time frame.
However, if a `constraint` block some state transition,
a property violation could be masked, leading to an incorrect unsat result.

I managed to construct a minimal example (where Pono returns incorrect unsat):

```
1 sort bitvec 1
2 const 1 1
3 state 1 s
4 init 1 3 2
5 next 1 3 -2
6 constraint 3
7 input 1 i
8 bad 7
```

The state `s` is initialized to 1, but remains 0 in subsequent time frames.
The `bad` signal solely depends on input `i`.
The `constraint` at line 6 effectively prevents any valid state transitions beyond the initial time frame.
However, the property can be violated at the initial (0th) time frame,
so model checkers should return "sat".

Po-Chun-Chien:
@ahmed-irfan @CyanoKobalamyne 
Is there a reason why a monitor is needed when the property depends on an input?
In fact, when running with the flag `--promote-inputvars` or if I changed line 7 to `7 state 1 i`,
Pono can correctly return sat.
(A `state` without `next` behaves the same as `input` in Btor2.)

ahmed-irfan:
[...]

PR #379, opened 2025-03-27
Promote inputs to states in Btor2's bad

Po-Chun-Chien:
This PR fixes the wrong proof in #377.

However, the fix applies only to Btor2 models.
The soundness issue may still occur with other models (see https://github.com/stanford-centaur/pono/issues/377#issuecomment-2759013003).


ahmed-irfan:
Can you also add the test from #377 ?

Po-Chun-Chien:
> Can you also add the test from #377 ?

Done :white_check_mark:

PR #381, opened 2025-03-31
New options and improvements to interpolation-based engine

Po-Chun-Chien:
This PR introduces several configurable options for Pono's `interp` engine:

- Frontier-set simplification (default: enabled)
- Time frames at which the property is considered during interpolation (default: all)
- Eager unrolling (default: disabled)
- Direction of interpolant computation (default: forward)
- ~~cvc5 as a backend solver~~ (moved to #385)

An evaluation was conducted at commit b97fc71 across 395 benchmark tasks,
with each configuration demonstrating different strengths in the [results](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-251-gb97fc71/pono-interp.table.html).


engines/interpolantmc.cpp
```diff
@@ -120,26 +124,42 @@ bool InterpolantMC::step(int i)
   }
 
   Term bad_i = unroller_.at_time(bad_, i);
-  bad_disjuncts_ = solver_->make_term(Or, bad_disjuncts_, bad_i);
+  if (interp_props_ == InterpPropsEnum::ONLY_LAST || i == 1) {
```
ahmed-irfan, line 127: first should be i == 0

engines/interpolantmc.cpp
```diff
@@ -120,26 +124,42 @@ bool InterpolantMC::step(int i)
...
+    // Note that by skipping the property at the first time frame,
+    // the computed interpolant might contain bad states
+    bad_disjuncts_ = bad_i;
+  } else if (interp_props_ == InterpPropsEnum::FIRST_AND_LAST) {
+    // Only take the first and last properties; the rest are skipped.
+    bad_disjuncts_ = solver_->make_term(Or, unroller_.at_time(bad_, 1), bad_i);
```
ahmed-irfan, line 133: unroller_.at_time(bad_, 0) ?

options/options.cpp
```diff
@@ -587,6 +591,37 @@ const option::Descriptor usage[] = {
...
+    "  --no-interp-frontier-simpl \tDisable frontier-set simplification in "
+    "interp engine" },
+  { INTERP_PROPS,
+    0,
+    "",
+    "interp-props",
+    Arg::NonEmpty,
+    "  --interp-props \tSpecifies at which time frames properties are "
+    "considered when computing interpolants: all, first-and-last, only-last "
+    "(WARNING: choosing not 'all' could cause incompleteness on some "
```
ahmed-irfan, line 608: incompleteness in the case of QF_BV / QF_ABV benchmarks, right?
What about other theories?

ahmed-irfan: Also, what result is reported when using only-last?

options/options.cpp
```diff
@@ -587,6 +591,37 @@ const option::Descriptor usage[] = {
...
+    "interp-eager-unroll",
+    Arg::None,
+    "  --interp-eager-unroll \tUnroll the transition system eagerly in interp "
+    "engine" },
+  { INTERP_BACKWARD,
+    0,
+    "",
+    "interp-backward",
+    Arg::None,
+    "  --interp-backward \tCompute interpolants in a backward manner in interp "
```
ahmed-irfan, line 623: please add text about the other case when this option is not use. something like " by default interpolates are computed ..... "

options/options.cpp
```diff
@@ -811,6 +846,23 @@ ProverResult PonoOptions::parse_and_set_options(int argc,
...
+            interp_props_ = InterpPropsEnum::ALL;
+          } else if (opt.arg == std::string("first-and-last")) {
+            interp_props_ = InterpPropsEnum::FIRST_AND_LAST;
+          } else if (opt.arg == std::string("only-last")) {
+            interp_props_ = InterpPropsEnum::ONLY_LAST;
+          } else {
+            throw PonoException("Unknown --interp-props option: "
+                                + std::string(opt.arg));
+            break;
+          }
```
ahmed-irfan, line 863: missing a break here.

engines/interpolantmc.cpp
```diff
@@ -178,8 +198,18 @@ bool InterpolantMC::step(int i)
   // transB can't have any symbols from time 0 in it
   assert(i > 0);
   // extend the unrolling
-  transB_ = solver_->make_term(And, transB_, unroller_.at_time(ts_.trans(), i));
-  ++reached_k_;
+  if (unroll_eagerly_) {
```
ahmed-irfan, line 201: please add more comments about eager unrolling

ahmed-irfan: can you add a reference to this?

ahmed-irfan (commented):
see comments.

Also, can you please add some tests that exercise the new options.

engines/interpolantmc.cpp
Po-Chun-Chien: The initial/zeroth time step (i == 0) is handled by `step_0()`:

```c
  if (i == 0) {
    // Can't get an interpolant at bound 0
    // only checking for trivial bug
    return step_0();
  }
```

The "first" here refers to the states reached after "one" step from the initial state.

engines/interpolantmc.cpp
Po-Chun-Chien: 1 is correct (see also https://github.com/stanford-centaur/pono/pull/381#discussion_r2125858854)

options/options.cpp
Po-Chun-Chien: > incompleteness in the case of QF_BV / QF_ABV benchmarks, right?
What about other theories?

The algorithm is proven to be complete for finite-state systems.
For infinite-state systems, I don't think it is not complete.

> Also, what result is reported when using only-last?

The result (sat/unsat) is correct.
However, the derived fixed point (invariant) might contain unreachable bad states.
The reason is that if we do not put `~P(state@1)` in formula `B` for interpolation.
The computed interpolant might intersect with `~P`.

Note that if a bad state is reachable,
then a fixed point will not be derived and the bad state will be found by the BMC query.
But perhaps having bad states in the fixed point is undesirable for users,
we could also just remove this option.

options/options.cpp
Po-Chun-Chien: See de8a393

engines/interpolantmc.cpp
Po-Chun-Chien: See a more detailed explanation in 9369be3.
There is no reference AFAIK.

options/options.cpp
Po-Chun-Chien: Fixed in 6df742c

engines/interpolantmc.cpp
```diff
@@ -191,7 +224,7 @@ bool InterpolantMC::step_0()
 
   Result r = solver_->check_sat();
   if (r.is_unsat()) {
-    ++reached_k_;
+    reached_k_ = 0;
```
ahmed-irfan, line 222: why this change?

We maintain in reached_k_ (in all engines, I think) the bound we have proven to be unreachable.

options/options.cpp
ahmed-irfan: I think we should produce a fix point that contains bad states -- this causes confusion because we cannot say that to be an invariant strong enough to be a proof witness.

I think we should remove that case.

engines/interpolantmc.cpp
Po-Chun-Chien: `reached_k_` was `-1` before entering `step_0`,
so `++reached_k_` and `reached_k_ = 0` are equivalent here.

I slightly prefer the latter because it is clearer what `reached_k_`'s value is after this step,
but I don’t have a strong preference and can revert the change if requested.

options/options.cpp
Po-Chun-Chien: Done in c82afe0

Po-Chun-Chien (commented):
> remove 'all' from the option --interp-props

I suppose you meant "only-last", since "all" is the original behavior.
The option was removed in c82afe0.

(Sorry for the delay. I was occupied with other tasks.)

PR #384, opened 2025-04-02
Rename Btor2 symbols internally

### LINKED ISSUE #376: Exception caused by invalid symbol in Btor2
When running the IC3sa engine of Pono on this Btor2 circuit ([even.btor2](https://gitlab.com/sosy-lab/research/data/sv-benchmarks-translation/-/blob/b404abc8f1ce5436ab1dd0532ce6a8e932f47a68/translated/btor2/loop-invariants/even.btor2)):

```bash
./pono --engine ic3sa even.btor2
```

Pono (at commit 87d4a9e2) failed with the following exception:

```
Caught generic exception...
invalid symbol '|pc[2].next|.next', symbol is not SMT-LIB compliant
error
```

In fact, Pono's IC3sa failed on almost all the tasks in this [benchmark set](https://gitlab.com/sosy-lab/research/data/sv-benchmarks-translation).

(If I run another engine such as `ind`, the exception does not occur.)

Po-Chun-Chien:
This PR fixes #376.
Now Btor2 symbols are renamed internally into `{state,input,output,term}<id>`,
where `<id>` is the ID in Btor2.
A mapping between the internally-assigned names to the original names is stored,
and used to restore the signal names during witness printing.

This PR also closes #361 and #369, because they are no longer needed.

### DIFF
--- a/frontends/btor2_encoder.cpp
+++ b/frontends/btor2_encoder.cpp
@@ -253,7 +253,8 @@ void BTOR2Encoder::parse(const std::string filename)
   std::unordered_map<int64_t, uint64_t> id2statenum;
   smt::Sort linesort;
   smt::TermVec termargs;
-  std::string symbol;
+  std::string orig_symbol;
+  std::string new_symbol;
 
   Btor2LineIterator bt2_it = btor2parser_iter_init(reader);
   while (Btor2Line * bt2_line = btor2parser_iter_next(&bt2_it)) {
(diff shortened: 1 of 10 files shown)


Po-Chun-Chien:
@CyanoKobalamyne I reordered the commits.
The refactor commit is now the last one.
You can now review the changes more easily by checking the first 4 commits (534eaf8..dc00347).
 (Unfortunately, I couldn't place the refactor commit first,
 as it depends on the initial commit 534eaf8.)

CyanoKobalamyne (approved):

PR #385, opened 2025-04-02
Enable cvc5's interpolation

Po-Chun-Chien:
Smt-Switch already supports cvc5's interpolation API.
This PR simply enables this in Pono.

### DIFF
--- a/smt/available_solvers.cpp
+++ b/smt/available_solvers.cpp
@@ -231,6 +231,9 @@ SmtSolver create_interpolating_solver_for(SolverEnum se, Engine e)
 {
   if (ic3_variants().find(e) == ic3_variants().end()) {
     return create_interpolating_solver(se);
+  } else if (se != MSAT) {
+    throw SmtException(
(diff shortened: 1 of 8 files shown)


smt/available_solvers.cpp
```diff
@@ -225,6 +231,9 @@ SmtSolver create_interpolating_solver_for(SolverEnum se, Engine e)
 {
   if (ic3_variants().find(e) == ic3_variants().end()) {
     return create_interpolating_solver(se);
+  } else if (se != MSAT) {
+    throw SmtException(
+        "Running IC3-based engine with interpolation requires MathSAT.");
```
CyanoKobalamyne, line 236: Why did you add this? It should also work with cvc5, right?

Po-Chun-Chien: True, I've fixed it in d4aa6b052231212a0e5aadd8f078fdd1bd65236b

CyanoKobalamyne (approved):

PR #395, opened 2025-04-24
Clean up preprocessor directives and try-catch related to MathSAT

Po-Chun-Chien:
Since #385, MathSAT is not strictly required to run engines that use interpolation.
This PR removes outdated preprocessor directives and a try-catch block.

### DIFF
--- a/tests/test_utils.cpp
+++ b/tests/test_utils.cpp
@@ -134,8 +134,8 @@ TEST_P(UtilsEngineUnitTests, MakeProver)
   SolverEnum se = get<0>(GetParam());
   Engine eng = get<1>(GetParam());
 
-  if (eng == INTERP && se != MSAT) {
-    // skip interpolation unless the solver is MathSAT
+  if ((eng == INTERP || eng == IC3IA_ENGINE) && se != MSAT) {
+    // skip interpolation and IC3ia unless the solver is MathSAT
     return;
   }
 
--- a/engines/interpolantmc.cpp
+++ b/engines/interpolantmc.cpp
@@ -49,12 +49,12 @@ void InterpolantMC::initialize()
 
   super::initialize();
 
-  reset_assertions(interpolator_);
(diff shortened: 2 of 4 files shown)


tests/test_utils.cpp
```diff
@@ -134,8 +134,8 @@ TEST_P(UtilsEngineUnitTests, MakeProver)
   SolverEnum se = get<0>(GetParam());
   Engine eng = get<1>(GetParam());
 
-  if (eng == INTERP && se != MSAT) {
-    // skip interpolation unless the solver is MathSAT
+  if ((eng == INTERP || eng == IC3IA_ENGINE) && se != MSAT) {
+    // skip interpolation and IC3ia unless the solver is MathSAT
```
CyanoKobalamyne, line 138: I don't totally understand this. Shouldn't this test also work with cvc5 (or any other solver that supports interpolation)?

Po-Chun-Chien: The test will fail on any combination of the following options:

- `--smt-solver {bzla,cvc5}`
- `--smt-interpolator cvc5`
- `--engine {interp, ic3ia}`

The test case written in Btor2:

```
1 sort bitvec 8
2 sort bitvec 1
3 constd 1 1
4 constd 1 8
5 constd 1 0
6 state 1 x
7 init 1 6 5
8 add 1 6 3
9 next 1 6 8
10 ult 2 6 4
11 bad -10
```

Here's a summary of failing reasons for `(solver, interpolator, engine)`:

- `(bvla, cvc5, interp)`: Bitwuzla does not currently support `reset_assertions` (error)
- `(cvc5, cvc5, interp)`: Interpolant generation failed at some step (error)
- `(bvla, cvc5, ic3ia)`: Refinement failed at some step (unknown)
- `(cvc5, cvc5, interp)`: Expected term with sort `Bool` (error)

CyanoKobalamyne (approved):

PR #399, opened 2025-04-30
Cast init term to bit-vector

Po-Chun-Chien:
In Btor2, a bit-vector state can be initialized using a term that is the result of some operation.
However, if the initialization term is of boolean type, this can lead to issues during term construction due to the mismatching types.
To avoid this, the initialization term should always be cast to a bit-vector.

### DIFF
--- a/tests/encoders/test_btor2.cpp
+++ b/tests/encoders/test_btor2.cpp
@@ -119,6 +119,22 @@ TEST_P(Btor2UnitTests, InvalidSmtlibSymbol)
   ASSERT_NE(r, ProverResult::ERROR);
 }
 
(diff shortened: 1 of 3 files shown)


tests/encoders/test_btor2.cpp
```diff
@@ -119,6 +119,22 @@ TEST_P(Btor2UnitTests, InvalidSmtlibSymbol)
   ASSERT_NE(r, ProverResult::ERROR);
 }
 
+TEST_P(Btor2UnitTests, InitStateWithBool)
+{
+  // test BTOR2 file with invalid SMT-LIB symbol
```
CyanoKobalamyne:
```suggestion
  // test BTOR2 file with bool init value
```

PR #400, opened 2025-04-30
Add interpolation-sequence-based model checking algorithm

Po-Chun-Chien:
This PR implements interpolation-sequence-based model checking algorithm ([Vizel & Grumberg, FMCAD 2009](https://doi.org/10.1109/FMCAD.2009.5351148)).

A preliminary evaluation was conducted on 395 bit-vector tasks.
The [results](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-265-gc14daa6/pono-ismc.table.html#/) confirmed that the implementation is working and
showed that ISMC is able to solve some tasks that the other interpolation-based engines cannot.

In addition, I looked into some of the unexpected errors in the table.

- [`ERROR (2)`](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-265-gc14daa6/pono-ismc.table.html#/table?filter=2(0*status*(status(in(ERROR%20%282%29)),category(notIn())))): fixed by #399
- [SEGMENTATION FAULT](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-265-gc14daa6/pono-ismc.table.html#/table?filter=2(0*status*(status(in(SEGMENTATION%20FAULT)),category(notIn())))): cause by `msat::dpll::proof::TheoryLemma::~TheoryLemma()` (seems to be triggered by `msat_reset_env ()`; this [case](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-295-ga55bb6b/itp-engines.msat-incr.table.html#/table?hidden4=2,3,4&hidden9=2,3,4&hidden=0,1,2,3,5,6,7,8&filter=4(0*status*(status(in(SEGMENTATION%20FAULT)),category(notIn())))) can somehow be avoided by https://github.com/stanford-centaur/smt-switch/pull/411)


engines/interp_seq_mc.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
```
ahmed-irfan, line 19: add copyright header and reference of the paper. e.g. see https://github.com/stanford-centaur/pono/blob/main/engines/ic3ia.h

engines/interp_seq_mc.cpp
```diff
@@ -0,0 +1,245 @@
+#include "engines/interp_seq_mc.h"
```
ahmed-irfan, line 19: copyright header

Po-Chun-Chien: Done in e38267b

engines/interp_seq_mc.h
Po-Chun-Chien: Done in e38267b

engines/interp_seq_mc.cpp
```diff
@@ -0,0 +1,263 @@
...
+{
+  if (initialized_) {
+    return;
+  }
+
+  super::initialize();
+
+  interpolator_->reset_assertions();
+
+  // register UF symbols
```
ahmed-irfan: minor: you meant uninterpreted symbols

Po-Chun-Chien: UF should be correct here.
These symbols were introduced during abstraction (e.g., ceg_proph or ceg_bv_ops).
I expanded the code comments a bit.

Po-Chun-Chien:
> one more comment: do you think reset_assertions frequently effects the overall performance?

`.reset_assertions()` is called before checking entailment (`A => B`).
 These entailment checks are typically easy to solve and do not pose a performance bottleneck.

The interpolation queries are more heavy-lifting,
and are handled by a different solver instance.
With https://github.com/stanford-centaur/smt-switch/pull/411,
these queries can be done incrementally without resetting assertions.

> Can we do better?

I don't have a concrete idea at the moment.
I'll go ahead and merge this for now, and open a new PR if we come up with a better approach later.

PR #407, opened 2025-06-04
Assign unique artifact name to each job

Po-Chun-Chien:
Since #401, there could be multiple failing tests which upload test logs as artifacts.

According to the [README](https://github.com/actions/upload-artifact?tab=readme-ov-file#not-uploading-to-the-same-artifact) of `@actions/upload-artifact`,
we should assign a unique artifact name to each job.


CyanoKobalamyne (approved):
Thanks!

PR #415, opened 2025-06-26
Create separate safety and liveness property classes

CyanoKobalamyne:
In preparation for creating separate proper classes for safety and liveness properties..


CyanoKobalamyne:
Yes, I saw that. But I don't think we need subclasses here, since they are so similar. This also avoids breaking any existing usage of the class.

Po-Chun-Chien:
One advantage of using subclasses is type safety.
Once we introduce safety and liveness provers, which handle different *types* of properties,
subclasses ensure that the correct property type is passed to the prover at *compile time*.

If we rely on checking property labels within the prover, the validation can only occur at runtime.

That said, I agree that using labels requires fewer changes to the existing codebase.

I'll leave the final decision to you.

PR #418, opened 2025-06-28
Create BaseProver and SafetyProver classes

CyanoKobalamyne:


engines/ceg_prophecy_arrays.cpp
Po-Chun-Chien, line 1: There are some deprecated uses of `Property` and `Prover` in this file:

https://github.com/stanford-centaur/pono/blob/66d7f9fdb572a84dc61ccf10130b37ff6f783c64/engines/ceg_prophecy_arrays.cpp#L94

https://github.com/stanford-centaur/pono/blob/66d7f9fdb572a84dc61ccf10130b37ff6f783c64/engines/ceg_prophecy_arrays.cpp#L98

engines/msat_ic3ia.h
Po-Chun-Chien, line 1: CPP files and tests have not been updated:

https://github.com/stanford-centaur/pono/blob/66d7f9fdb572a84dc61ccf10130b37ff6f783c64/engines/msat_ic3ia.cpp#L34

https://github.com/stanford-centaur/pono/blob/66d7f9fdb572a84dc61ccf10130b37ff6f783c64/tests/test_msat_ic3ia.cpp#L40

https://github.com/stanford-centaur/pono/blob/66d7f9fdb572a84dc61ccf10130b37ff6f783c64/tests/test_msat_ic3ia.cpp#L64

utils/make_provers.cpp
```diff
@@ -43,11 +43,11 @@ vector<Engine> all_engines()
   return { BMC, BMC_SP, KIND, MBIC3, INTERP, IC3IA_ENGINE, IC3SA_ENGINE };
 }
 
-shared_ptr<Prover> make_prover(Engine e,
```
Po-Chun-Chien, line 46: We could give the functions more explicit names: `make_*_safety_prover`

engines/prover.h
```diff
@@ -86,47 +59,73 @@ class Prover
...
+  SafetyProver(const SafetyProperty & p,
+               const TransitionSystem & ts,
+               const smt::SmtSolver & s,
+               PonoOptions opt = PonoOptions());
 
-  int reached_k_;  ///< the last bound reached with no counterexamples
+  void initialize() override;
 
-  smt::Term bad_;
+  virtual bool witness(std::vector<smt::UnorderedTermMap> & out);
```
Po-Chun-Chien, line 91: This could be moved to `BaseProver` class, as provers for other properties should also be able to produce witnesses.

engines/prover.cpp
```diff
@@ -31,49 +31,85 @@ using namespace smt;
...
 {
   if (initialized_) {
     return;
   }
 
-  reached_k_ = -1;
-
   if (!ts_.only_curr(bad_)) {
     throw PonoException(
         "Property should not contain inputs or next state variables");
```
Po-Chun-Chien:
```suggestion
        "Safety property should not contain inputs or next state variables");
```

utils/make_provers.cpp
CyanoKobalamyne: I think leaving them like this is fine, because we can just have more overloads with `LivenessProperty` to make liveness provers.

engines/prover.h
CyanoKobalamyne: This is already done in #419.

PR #419, opened 2025-07-01
Create liveness prover class

CyanoKobalamyne:


Po-Chun-Chien:
In the `BaseProver` class, there's a field  `engine_`, which is of an enum type `Engine`.

Do you plan to (1) have two separate enums for safety and liveness engines or (2) list all engines (both liveness and safety) in `Engine`?
In case of (1), the field should be moved to `SafetyProver` class.

engines/prover.cpp
```diff
@@ -26,18 +27,19 @@
...
 
 namespace pono {
+
 BaseProver::BaseProver(const TransitionSystem & ts, const smt::SmtSolver & s)
     : BaseProver(ts, s, PonoOptions())
 {
 }
 
 BaseProver::BaseProver(const TransitionSystem & ts,
-                       const smt::SmtSolver & s,
```
Po-Chun-Chien, line 40: Maybe also remove another `smt::` several lines above?

engines/prover.cpp
```diff
@@ -90,45 +92,16 @@ Term BaseProver::to_orig_ts(Term t)
...
-  if (!ts_.only_curr(bad_)) {
-    throw PonoException(
-        "Safety property should not contain inputs or next state variables");
-  }
-
-  BaseProver::initialize();
-}
-
-bool SafetyProver::witness(std::vector<UnorderedTermMap> & out)
+bool BaseProver::witness(std::vector<UnorderedTermMap> & out)
```
Po-Chun-Chien: The default implementation for `witness()` and `witness_length()` are more tailored for `SafetyProver`.

I would prefer keeping these two APIs purely virtual at `BaseProver`,
and move the implementations to the `SafetyProver` class.

For the `LivenessProver` class, we can keep them virtual or throw NYI exceptions.

Po-Chun-Chien: If we later find out the implementations here can be shared between two prover classes,
we can refactor.

Po-Chun-Chien: After a second look, `BaseProver::witness` seems fine as it is only transferring the terms.
`witness_` is populated by `SafetyProver::compute_witness()`.

However, I am still not sure if it's a good idea to let `witness_length()` return `reached_k_ + 1` by default for `LivenessProver`.

engines/prover.cpp
CyanoKobalamyne: Done.

CyanoKobalamyne:
> In the `BaseProver` class, there's a field `engine_`, which is of an enum type `Engine`.
> 
> Do you plan to (1) have two separate enums for safety and liveness engines or (2) list all engines (both liveness and safety) in `Engine`? In case of (1), the field should be moved to `SafetyProver` class.

I was planning to add all engines to the same enum. We could separate it into two, but that would require duplicating a lot of logic, and we already have runtime checks for the value of `engine_`.

engines/prover.cpp
CyanoKobalamyne: You're right, we should move both the `reached_k_` field and the implementation of `witness_length()` into `SafetyProver`, because it doesn't really make sense for all liveness algorithms.

CyanoKobalamyne: Done.

PR #421, opened 2025-07-02
Export violation witnesses for IC3 engines

Po-Chun-Chien:
IC3 engines stores counterexample path in `cex_`,
which, AFAICS, only contains values of state variables.
We can use this information and do a BMC check to get the input values at each time step.

In this PR, a simple solution as describe above is implemented.
In later PRs, we could optimize the code to avoid recomputation.
(For example, in ic3sa and ic3ia, such BMC checks are already performed during refinement.)

### DIFF
--- a/engines/ic3base.cpp
+++ b/engines/ic3base.cpp
@@ -175,6 +175,7 @@ ProverResult IC3Base::check_until(int k)
       } else if (s == REFINE_NONE) {
         // this is a real counterexample
         assert(cex_.size());
+        compute_witness();
         return ProverResult::FALSE;
       } else {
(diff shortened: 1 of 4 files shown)


Po-Chun-Chien:
@luiza778 Could you test this PR on some Btor2 benchmarks?

Po-Chun-Chien:
@luiza778 kindly helped test the witness export functionality of `ic3sa` and `ic3bits` on over 200 Btor2 instances (many thanks!) No issues have been found so far.

@CyanoKobalamyne Could you please take a look at this PR along with #422?

CyanoKobalamyne (approved):

PR #423, opened 2025-07-03
Replace Property with SafetyProperty in engines and tests

Po-Chun-Chien:
This PR follows up on #415 and replaces remaining instances of the deprecated usage of `Property`.


Po-Chun-Chien:
Just notice the updates here have been covered by #418

PR #424, opened 2025-07-03
Convert size_t options to unsigned long

CyanoKobalamyne:
Necessitated by the lack of standard string conversion functions to `size_t`, see #406.

We also don't need any of these to be `size_t`; `unsigned long` is perfectly fine.


ahmed-irfan (approved):
LGTM

PR #426, opened 2025-07-07
Extend BV arith CEGAR to support more backend engines

Po-Chun-Chien:
This PR integrates more backend engines (including `bmc`, `ind`, and `interp`) into the CEGAR prover that abstracts bit-vector arithmetic.
To achieve this, the engine must implement the `reset_env()` API,
which resets its internal state such that it can start another verification run on the refined transition system.

Here are the evaluation results of [`ind`](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-282-g7a11d4f/kind.cegbv.table.html), [`interp`](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-282-g7a11d4f/interp.cegbv.table.html), [`bmc`](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-282-g7a11d4f/bmc.cegbv.table.html), and [`bmc-sp`](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-282-g7a11d4f/bmc-sp.cegbv.table.html).
With the abstraction, both [`ind`](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-282-g7a11d4f/kind.cegbv.table.html#/table?filter=0(0*status*(status(notIn()),category(in(error)))),2(0*status*(status(notIn()),category(in(correct))))) and [`interp`](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-282-g7a11d4f/interp.cegbv.table.html#/table?filter=0(0*status*(status(notIn()),category(notIn(correct)))),2(0*status*(status(notIn()),category(in(correct))))) were able to solve several Btor2 instances that were previously unsolvable without it.

### DIFF
--- a/engines/ic3base.cpp
+++ b/engines/ic3base.cpp
@@ -183,6 +183,7 @@ ProverResult IC3Base::check_until(int k)
       } else if (s == REFINE_NONE) {
         // this is a real counterexample
         assert(cex_.size());
+        compute_witness();
         return ProverResult::FALSE;
       } else {
         assert(s == REFINE_FAIL);
@@ -201,11 +202,6 @@ ProverResult IC3Base::check_until(int k)
   return ProverResult::UNKNOWN;
 }
 
-bool IC3Base::witness(std::vector<smt::UnorderedTermMap> & out)
-{
-  throw PonoException("IC3 witness NYI");
-}
-
 size_t IC3Base::witness_length() const
 {
   // expecting there to have been a witness computed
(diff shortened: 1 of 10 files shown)


Po-Chun-Chien:
I noticed that some [segmentation faults](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-282-g7a11d4f/interp.cegbv.table.html#/table?filter=2(0*status*(status(in(SEGMENTATION%20FAULT)),category(notIn())))) occur when running `interp` in CEGAR.
Based on preliminary investigation, the errors appear to originate from within MathSAT rather than from the changes introduced in this PR.

This issue will be investigated separately and is not part of this PR.
Running Pono with an smt-switch version that supports incremental interpolation queries avoids some of these errors (see https://github.com/stanford-centaur/smt-switch/pull/411#issuecomment-3050522546).

engines/cegar_ops_uf.cpp
```diff
@@ -252,7 +249,21 @@ void CegarOpsUf<IC3IA>::refine_subprover_ts(const UnorderedTermSet & axioms,
...
+{
+  refine_subprover_ts_base(axioms, skip_init);
+  super::reset_env();
+}
+
+template <>
+void CegarOpsUf<IC3IA>::refine_subprover_ts(const UnorderedTermSet & axioms,
+                                            bool skip_init)
+{
+  refine_subprover_ts_base(axioms, skip_init);
```
ahmed-irfan, line 266: what's the difference between the two refine methods?

tests/test_cegar_ops_uf.cpp
```diff
@@ -1,113 +1,143 @@
...
-  RelationalTransitionSystem rts = counter_ts(s, x);
-  Term prop_term = rts.make_term(Lt, x, rts.make_term(11, sort));
-  SafetyProperty prop(s, prop_term);
-
-  shared_ptr<CegarOpsUf<IC3IA>> ceg =
-      make_shared<CegarOpsUf<IC3IA>>(prop, rts, s);
-  ceg->set_ops_to_abstract({ Plus });
-
-  ProverResult r = ceg->check_until(5);
+  if (opts.engine_ == Engine::BMC || opts.engine_ == Engine::BMC_SP) {
```
ahmed-irfan: maybe run them and check for UNKNOWN result

engines/cegar_ops_uf.cpp
Po-Chun-Chien: `refine_subprover_ts_base` performs common refinement steps that can be shared across multiple prover classes, so I refactored it into a separate function. This function only adds refining axioms to `prover_ts_` and does not modify other internal states of the prover.

The prover can either perform a cold restart (via `.reset_env()` introduced in this PR), which resets the solver’s state, or a warm restart, which preserves (part of) its previous state (e.g., `IC3IA::reabstract()`). The latter integrates more tightly with the CEGAR flow but requires more implementation effort.

tests/test_cegar_ops_uf.cpp
Po-Chun-Chien: see df533eb

engines/cegar_ops_uf.cpp
ahmed-irfan: Oh, I see. At first, the body of these two methods looked same to me ;-)

PR #428, opened 2025-07-07
Abstract bit-vector arithmetic using free state symbols

Po-Chun-Chien:
This PR introduces an alternative approach for abstracting bit-vector arithmetic operations using free state symbols.
Compared to using uninterpreted functions, this abstraction is coarser (less precise).
The CEGAR implementation can be shared between the two abstraction methods.

The table below summarizes the number of solved tasks on a benchmark set of IC3IA/IC3SA with and without abstraction ([more info](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-274-g5203f52/cegbv.uf-vs-free-var.table.html)):

|       | cegbv (free var) | default | diff    | cegbv (uf) | diff    |
|-------|------------------|---------|---------|------------|---------|
| IC3IA |              751 |     735 | +30/-14 |        750 | +14/-13 |
| IC3SA |              731 |     732 | +21/-22 |        736 | +15/-20 |

The results indicate that this new abstraction is able to solve certain tasks that the previous approach could not.

### DIFF
--- a/modifiers/ops_abstractor.h
+++ b/modifiers/ops_abstractor.h
@@ -26,7 +26,7 @@ class OpsAbstractor : public Abstractor
  public:
   OpsAbstractor(const TransitionSystem & conc_ts,
                 TransitionSystem & abs_ts,
-                bool abs_using_free_symbols);
+                bool abs_using_free_symbols = false);
 
   typedef Abstractor super;
 
--- a/modifiers/ops_abstractor.h
+++ b/modifiers/ops_abstractor.h
@@ -24,7 +24,9 @@ namespace pono {
 class OpsAbstractor : public Abstractor
 {
  public:
-  OpsAbstractor(const TransitionSystem & conc_ts, TransitionSystem & abs_ts);
+  OpsAbstractor(const TransitionSystem & conc_ts,
+                TransitionSystem & abs_ts,
+                bool abs_using_free_symbols);
 
   typedef Abstractor super;
 
@@ -78,6 +80,8 @@ class OpsAbstractor : public Abstractor
 
   smt::UnorderedTermMap abs_terms_;  // abs uf to concrete operator -- only
                                      // replace the top-level uf
+
+  const bool abs_using_free_symbols_;
 };
 
 }  // namespace pono
--- a/options/options.cpp
+++ b/options/options.cpp
@@ -461,8 +461,9 @@ const option::Descriptor usage[] = {
     "",
     "ceg-bv-arith-as-free-symbol",
     Arg::None,
-    "  --ceg-bv-arith-as-free-symbol \tabstractBV arithmetic operators using "
-    "free symbols (unconstrained inputs) instead of uninterpreted functions" },
+    "  --ceg-bv-arith-as-free-symbol \tabstract BV arithmetic operators using "
+    "free symbols (unconstrained state variables) instead of uninterpreted "
+    "functions" },
   { CEG_BV_ARITH_MIN_BW,
     0,
     "",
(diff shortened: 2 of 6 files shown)


Po-Chun-Chien:
I will add tests for the new abstraction after #426 is merged in order to reuse the test suite introduced in 
66c6663.

Po-Chun-Chien:
I crafted a small example ([abs.btor2.gz](https://github.com/user-attachments/files/21398613/abs.btor2.gz)) to see if the abstraction is done correctly in `init` and `bad`.

- Concrete transition system and property:

```
Init: (and true (= state6 (bvmul state3 state4)))
Trans: (and true (= state6.pono_generated__next (bvmul state6 state6)))
Bad: (not (not (= (bvcomp (bvmul state6 state6) #b00110001) #b1)))
```

- Abstraction with UF

```
Abs Init: (and true (= state6 (abs_bvmul_8_8_8 state3 state4)))
Abs Trans: (and true (= state6.pono_generated__next (abs_bvmul_8_8_8 state6 state6)))
Abs Bad: (not (not (= (bvcomp (abs_bvmul_8_8_8 state6 state6) #b00110001) #b1)))
```

(`abs_bvmul_8_8_8` is the introduced UF symbol)

- Abstraction with free state variable:

```
Abs Init: (and true (= state6 abs_bvmul_8_8_8_id1))
Abs Trans: (and true (= state6.pono_generated__next abs_bvmul_8_8_8_id0))
Abs Bad: (not (not (= (bvcomp abs_bvmul_8_8_8_id0 #b00110001) #b1)))
```

(`abs_bvmul_8_8_8_id0` and `abs_bvmul_8_8_8_id1` are the introduced free state symbols.)


AFAICS, this should be correct.

options/options.cpp
```diff
@@ -454,7 +455,14 @@ const option::Descriptor usage[] = {
...
     "  --ceg-bv-arith \tabstraction-refinement for the BV arithmetic operators "
-    "(mul, div, rem, mod)" },
+    "(mul, div, rem, mod) using uninterpreted functions." },
+  { CEG_BV_ARITH_AS_FREE_SYMBOL,
+    0,
+    "",
+    "ceg-bv-arith-as-free-symbol",
+    Arg::None,
+    "  --ceg-bv-arith-as-free-symbol \tabstractBV arithmetic operators using "
+    "free symbols (unconstrained inputs) instead of uninterpreted functions" },
```
ahmed-irfan: they are not inputs in the sense of Pono transition system. They are state variables (without next constraint)

modifiers/ops_abstractor.h
```diff
@@ -24,7 +24,9 @@ namespace pono {
 class OpsAbstractor : public Abstractor
 {
  public:
-  OpsAbstractor(const TransitionSystem & conc_ts, TransitionSystem & abs_ts);
+  OpsAbstractor(const TransitionSystem & conc_ts,
+                TransitionSystem & abs_ts,
+                bool abs_using_free_symbols);
```
CyanoKobalamyne:
```suggestion
                bool abs_using_free_symbols = false);
```
This is a public API, so we should provide a default argument here to not break backwards-compatibility.

CyanoKobalamyne (changes requested):
Looks good, modulo the two pending comments.

options/options.cpp
Po-Chun-Chien: see bab13dd

CyanoKobalamyne:
The failures are due to an outage GitHub is experiencing right now: https://www.githubstatus.com/

PR #433, opened 2025-07-22
Unroll formulas at step 0 of ISMC

Po-Chun-Chien:
At step 0 of ISMC, we check the satisfiability of $Init(s) \wedge \neg P(s)$.
The untimed and timed versions of the formula are equisatisfiable.
However, to compute the witness with `compute_witness()`, the variables have to be timed.


Po-Chun-Chien:
@luiza778 Could you test the updated ISMC engine in this PR?

luiza778:
Yes!

Po-Chun-Chien:
This has been tested on over 200 Btor2 instances (many thanks to @luiza778). All witnesses can be validated using BtorSim.

PR #434, opened 2025-07-28
Implement Dual Approximated Reachability

Po-Chun-Chien:
This PR introduces a new engine `dar` (Dual Approximated Reachability), an unbounded model checking algorithm proposed by Yakir Vizel, Orna Grumberg, and Sharon Shoham in the paper "[Intertwined Forward-Backward Reachability Analysis Using Interpolants](https://doi.org/10.1007/978-3-642-36742-7_22)", published in TACAS 2013.

A preliminary [evaluation](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-297-g589253b/itp-engine-cmp.table.html) was conducted on 395 bit-vector Btor2 tasks. In this benchmark, `dar` demonstrated performance comparable to other interpolation-based engines such as `interp` and `ismc`. A few [errors](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-297-g589253b/itp-engine-cmp.table.html#/table?filter=0(0*status*(status(in(ERROR%20%282%29)),category(notIn())))) were observed, which, upon investigation, were traced back to incorrect interpolation sequences computed by MathSAT.

Summary of evaluation:
|        | dar | interp | diff  | ismc | diff   |
|--------|-----|--------|-------|------|--------|
| #solved | 134 |    137 | +6/-9 |  123 | +18/-7 |

Note: In the current implementation, the solver is reset at each call to functions `*_strengthen`. Maintaining solver state between calls may improve performance, but this optimization is left for future work.


Po-Chun-Chien:
@luiza778 Could you check the witnesses produced by this new engine `dar`?

luiza778:
Yes!

Po-Chun-Chien:
Witness generation has been tested on over 200 Btor2 instances (many thanks to @luiza778). All witnesses can be validated using BtorSim.

engines/dual_approx_reach.cpp
```diff
@@ -0,0 +1,353 @@
...
+    solver_->assert_formula(solver_->make_term(And, f0, b1));
+    Result r = solver_->check_sat();
+    if (r.is_unsat()) {
+      // found an index such that the conjunction is unsat
+      break;
+    }
+    if (unsat_idx == seq_len - 1) {
+      // no such index found, return false
+      logger.log(1, "DAR: local strengthening failed");
+      return false;
```
ahmed-irfan, line 202: pop here as well -- just to be safe if in future we make don't do reset assertions

engines/dual_approx_reach.cpp
```diff
@@ -0,0 +1,353 @@
...
+      // found an index for refinement
+      int_assertions.push_back(to_interpolator_.transfer_term(ub));
+      break;
+    }
+    if (unsat_idx == seq_len - 1) {
+      // this is a concrete counterexample
+      logger.log(1, "DAR: Found a concrete CEX");
+      return false;
+    }
+    solver_->pop();  // pop backward_seq_ assertion
```
ahmed-irfan: only one pop... is this intentional?

engines/dual_approx_reach.cpp
```diff
@@ -0,0 +1,353 @@
...
+    Result r = solver_->check_sat();
+    if (r.is_unsat()) {
+      // found an index for refinement
+      int_assertions.push_back(to_interpolator_.transfer_term(ub));
+      break;
+    }
+    if (unsat_idx == seq_len - 1) {
+      // this is a concrete counterexample
+      logger.log(1, "DAR: Found a concrete CEX");
+      return false;
```
ahmed-irfan, line 252: add a pop here as well before returning false

ahmed-irfan (commented):
left some comments

engines/dual_approx_reach.cpp
Po-Chun-Chien: This is intended. The solver state should be kept such that the later procedure can extract the witness.
I added some explanation in c50b089.

engines/dual_approx_reach.cpp
Po-Chun-Chien: Same as https://github.com/stanford-centaur/pono/pull/434#discussion_r2267931791

engines/dual_approx_reach.cpp
Po-Chun-Chien: Yes, the unrolled transition relation is kept on the solver stack and will be reused in the next iteration.
I added some comments in c50b089.

ahmed-irfan (approved):

PR #438, opened 2025-08-12
Enable BV index sort in CEGP

Po-Chun-Chien:
This PR implements the "low-effort" approach for BV indices as described in the [CEGP paper](https://lmcs.episciences.org/9984/pdf):

> One low-effort approach for applying counterexample-guided prophecy over finite-domain indices is to give up completeness. By simply not including axioms over a λ index for finite-domain sorts, the array solving procedure might conclude the query is satisfiable when it is actually unsatisfiable. Thus, the overall algorithm could return spurious counterexamples, but would still soundly return proofs. A given spurious counterexample is finite and would be straightforward to analyze, either with a dedicated checker or an SMT solver without this limitation.

If no further refinement can be made, the results are havocked from `FALSE` to `UNKNOWN` to avoid spurious counterexamples.

A preliminary evaluation is done on 1521 Btor2 tasks with arrays.
The [results](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-290-g8527afd/cegp-bv-index.table.html) show that [`cegp+ic3ia`](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-290-g8527afd/cegp-bv-index.table.html#/table?hidden=2,3&hidden0=2,4&hidden1=2,4&hidden4=2,4&filter=0(0*status*(status(notIn()),category(notIn()))),1(0*status*(status(notIn()),category(in(correct)))),4(0*status*(status(notIn()),category(in(error))))) and [`cegp+interp`](https://www.cip.ifi.lmu.de/~chien/benchexec-results/pono/0.1.1-290-g8527afd/cegp-bv-index.table.html#/table?hidden=0,1&hidden0=2,4&hidden1=2,4&hidden4=2,4&hidden2=2,4&hidden3=2,4&filter=2(0*status*(status(notIn()),category(notIn()))),3(0*status*(status(notIn()),category(in(correct)))),4(0*status*(status(notIn()),category(in(error))))) could solve some tasks that `ind` could not.

Related to #271.

### DIFF
--- a/engines/ceg_prophecy_arrays.cpp
+++ b/engines/ceg_prophecy_arrays.cpp
@@ -10,7 +10,8 @@
 ** directory for licensing information.\endverbatim
 **
 ** \brief An implementation of Counter-Example Guided Prophecy for array
-**        model checking. It is parameterized by an underlying model checking
+**        model checking (https://lmcs.episciences.org/9984/pdf).
+**        It is parameterized by an underlying model checking
 **        procedure which need not handle arrays (only UF). However, a common
 **        instantiation is with an IC3-style procedure, in which case we
 **        often refer to this algorithm as "prophic3".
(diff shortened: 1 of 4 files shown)


engines/ceg_prophecy_arrays.cpp
```diff
@@ -225,18 +241,21 @@ void CegProphecyArrays<Prover_T>::initialize()
...
       contains_arrays = true;
       SortKind sk = sort->get_indexsort()->get_sort_kind();
-      if (sk != REAL && sk != INT) {
-        throw PonoException(
-            "CEGP currently only supports infinite domain indices in arrays "
-            "due to an edge case for constant arrays, but got "
-            + sort->to_string());
+      if (sk == REAL || sk == INT) {
+        // do nothing
+      } else if (sk == BV) {
```
ahmed-irfan: Add Bool here as well

Po-Chun-Chien: Done in 996bf0b

PR #440, opened 2025-08-14
Allow ISMC and DAR run in CEGAR BV arith

Po-Chun-Chien:
Similar to #426.

### DIFF
--- a/tests/test_cegar_ops_uf.cpp
+++ b/tests/test_cegar_ops_uf.cpp
@@ -14,9 +14,9 @@ unordered_set<Engine> get_cegar_ops_uf_engines()
 {
   return {
 #ifdef WITH_MSAT
-    IC3IA_ENGINE, INTERP,
+    IC3IA_ENGINE, INTERP, ISMC, DAR,
 #endif
-    IC3SA_ENGINE, BMC_SP, BMC, KIND
+    IC3SA_ENGINE, BMC_SP, BMC,  KIND
   };
 }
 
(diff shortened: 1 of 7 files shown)


Po-Chun-Chien:
@luiza778 Could you also help check the witnesses for the engines `ismc` and `dar`, following the same procedure as in https://github.com/stanford-centaur/pono/pull/436#issuecomment-3137415652?

tests/test_cegar_ops_uf.cpp
```diff
@@ -14,9 +14,9 @@ unordered_set<Engine> get_cegar_ops_uf_engines()
 {
   return {
 #ifdef WITH_MSAT
-    IC3IA_ENGINE, INTERP,
+    IC3IA_ENGINE, INTERP, ISMC, DAR,
 #endif
-    IC3SA_ENGINE, BMC_SP, BMC, KIND
+    IC3SA_ENGINE, BMC_SP, BMC,  KIND
```
CyanoKobalamyne, line 19: ?

tests/test_cegar_ops_uf.cpp
CyanoKobalamyne, line 1: What's the reason for changing the bounds for the BMC, BMC-SP, etc. tests here?

CyanoKobalamyne (commented):
Thanks! Left a couple of comments.

ahmed-irfan (commented):
Other than the comments from Aron, the PR LGTM

tests/test_cegar_ops_uf.cpp
Po-Chun-Chien: This is done automatically by the formatter.
I think it is trying to match the starting position of `DAR` at line above.

tests/test_cegar_ops_uf.cpp
Po-Chun-Chien: I increased the unrolling bounds for the safe cases so that ISMC and DAR can successfully derive proofs (BMC-SP can also prove the test cases safe at these higher bounds).

For the unsafe cases, I reduced the bounds from 11 to 10, and counterexamples can still be found.
The changes to the unsafe cases are optional and can be reverted if needed.

tests/test_cegar_ops_uf.cpp
CyanoKobalamyne: Oh okay. Then let's leave it like this.

tests/test_cegar_ops_uf.cpp
CyanoKobalamyne: Sounds good, just wanted to understand.

CyanoKobalamyne (approved):

PR #442, opened 2025-08-19
Add tests for CEX computation and the correctness of invariant in Ceg-Ops-Uf engines

Po-Chun-Chien:
This PR adds tests for CEX computation and invariant correctness in Ceg-Ops-Uf engines.
Similar to #422, the tests checks that CEXs can be successfully computed, but do not validate their correctness.
The PR also fixes a bug in invariant concretization.

### DIFF
--- a/engines/cegar_ops_uf.cpp
+++ b/engines/cegar_ops_uf.cpp
@@ -21,8 +21,10 @@
 #include "engines/bmc.h"
 #include "engines/bmc_simplepath.h"
 #include "engines/ceg_prophecy_arrays.h"
+#include "engines/dual_approx_reach.h"
 #include "engines/ic3ia.h"
 #include "engines/ic3sa.h"
+#include "engines/interp_seq_mc.h"
 #include "engines/interpolantmc.h"
 #include "engines/kinduction.h"
 #include "smt/available_solvers.h"
@@ -366,9 +368,11 @@ void CegarOpsUf<CegProphecyArrays<IC3IA>>::refine_subprover_ts(
 // TODO add other template classes
 template class CegarOpsUf<Bmc>;
 template class CegarOpsUf<BmcSimplePath>;
+template class CegarOpsUf<DualApproxReach>;
 template class CegarOpsUf<IC3IA>;
 template class CegarOpsUf<IC3SA>;
 template class CegarOpsUf<InterpolantMC>;
+template class CegarOpsUf<InterpSeqMC>;
 template class CegarOpsUf<KInduction>;
 template class CegarOpsUf<CegProphecyArrays<IC3IA>>;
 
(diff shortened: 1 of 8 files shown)


engines/cegar_ops_uf.cpp
```diff
@@ -83,13 +83,16 @@ ProverResult CegarOpsUf<Prover_T>::check_until(int k)
   }
 
   if (res == ProverResult::TRUE && super::invar_) {
-    // update the invariant
-
-    // TODO currently getting some errors in the concretization phase.
```
Po-Chun-Chien, line 90: @ahmed-irfan  Any chance you recall the errors and their cause?
I’ve fixed the bug in d8a6eb0 that broke the tests.

ahmed-irfan: I don't recall what was the issue.

modifiers/ops_abstractor.cpp
```diff
@@ -281,9 +281,23 @@ WalkerStepResult OpsAbstractor::ConcretizationWalker::visit_term(Term & term)
...
+          != oa_.abstraction_cache_.end()) {
+        // Corner case: the abstract term itself is an operation to abstract.
+        // This can occur in invariants derived from the prover.
+        // If the concrete term is already in the cache, calling
+        // `oa_.update_term_cache` triggers an assertion failure.
+        // In this case, only record the abstract-to-concrete term mapping.
+        assert(oa_.ops_to_abstract_.find(op) != oa_.ops_to_abstract_.end()
+               && (rebuilt->get_sort()->get_sort_kind() != BV
+                   || rebuilt->get_sort()->get_width() > oa_.min_bw_));
+        oa_.concretization_cache_[term] = rebuilt;
```
Po-Chun-Chien: https://github.com/stanford-centaur/pono/commit/d8a6eb0303c9c1f1dcb427686a67d01d85110d31#commitcomment-164274465 from @ahmed-irfan:

> how about we update the code of update_term_cache? If a term is present in the concretization/abstraction cache, then the mapped value in the cache should be equal to the one we are currently asking to update. What do you think?

Po-Chun-Chien: Sure, this could also be a way to go.

> If a term is present in the concretization/abstraction cache, then the mapped value in the cache should be equal to the one we are currently asking to update.

I wasn't sure if I should overwrite the concrete-to-abstract term mapping (`oa_.abstraction_cache_`) in this case, because then we would then lose the information of `(bvmul a b)` → `(abs_bvmul_uf a b)` (following the example [above](https://github.com/stanford-centaur/pono/pull/442#issuecomment-3206662468)) and the abstractor cannot be reused in the subsequent steps.

That's why I only opted for updating the abstract-to-concrete term mapping (`oa_.concretization_cache_`).
I could also move this part to `oa_.abstraction_cache_` as requested.

Po-Chun-Chien: @ahmed-irfan I refactored it in 4640748. What do you think?

Issue #441, opened 2025-08-14
Support witness export in parallel-execution script


Po-Chun-Chien:
In HWMCC 2025, exporting violation witnesses to a specified file is now mandatory.
Currently, the parallel-execution script does not provide this functionality.

(BTW, the latest version of the script exists only in the [hwmcc24-dev branch](https://github.com/stanford-centaur/pono/blob/hwmcc24-dev/scripts/parallel_pono.py). It should be merged into main to make it more accessible to users.)

To enable witness export, I recommend using the new `--dump-btor2-witness` option introduced in #373.

Po-Chun-Chien:
@CyanoKobalamyne would you take care of this?

PR #445, opened 2025-08-21
Simplify and reorganize parallel portfolio script

CyanoKobalamyne:
The engine configurations are the same here as those used at HWMCC 2024, except for k-induction with the invariant generalization turned on, because that is an experimental feature that has not been integrated into main.

The command line interface has been updated to match the one required for HWMCC 25:
```
<tool> <benchmark> <certificate.sat>
```

Addresses #441.

### DIFF
--- a/scripts/parallel_pono.py
+++ b/scripts/parallel_pono.py
@@ -96,7 +96,7 @@ def summarize(file: str, engine: str, returncode: int, runtime: float, cmd: list
 
 def clean_up(processes: dict[str, subprocess.Popen[str]], verbose: bool):
     for name, process in processes.items():
-        if process.poll() is None:
+        if process.poll() is None:  # process has not finished yet
             process.terminate()
         if verbose and process.stderr and (stderr := process.stderr.read()):
             for line in stderr.splitlines():
@@ -126,6 +126,7 @@ def main() -> int:
     start_times: dict[str, float] = {}
     atexit.register(clean_up, processes, args.verbose)
 
+    # Launch each portfolio solver as a subprocess.
     for name, options in ENGINE_OPTIONS.items():
         cmd = [executable, "-k", str(args.bound), *options]
         if args.witness_file:
@@ -141,10 +142,11 @@ def main() -> int:
         start_times[name] = time.time()
         processes[name] = proc
 
+    # Wait for at least one process to terminate successfully.
     while processes:
         for name, process in processes.items():
             end_time = time.time()
-            if process.poll() is not None:
+            if process.poll() is not None:  # process has finished
                 if args.summarize:
                     runtime = end_time - start_times[name]
                     cmd = cast(list[str], process.args)
--- a/scripts/parallel_pono.py
+++ b/scripts/parallel_pono.py
@@ -7,10 +7,12 @@ import csv
 import dataclasses
 import enum
 import logging
+import os
 import shutil
 import signal
 import subprocess
 import sys
+import tempfile
 import time
 from typing import cast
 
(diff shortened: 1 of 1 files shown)


scripts/parallel_pono.py
```diff
@@ -1,135 +1,133 @@
...
-                            print_process_output(bmc)
-                            shutdown = True
-                            break
+    # Try current directory if pono is not on PATH.
+    executable = shutil.which("pono") or "./pono"
+
+    for name, options in ENGINE_OPTIONS.items():
+        cmd = [executable, "-k", str(args.bound), *options]
+        if args.witness_file:
+            cmd.extend(["--dump-btor2-witness", args.witness_file])
```
Po-Chun-Chien: To avoid race condition, it is safer to write to a unique temp file first.
We can later move the witness to the specified location.

Also, remember to clean up the temp files as required by HWMCC.

Po-Chun-Chien (changes requested):

scripts/parallel_pono.py
```diff
@@ -1,135 +1,186 @@
...
+        with open(args.summarize, "w") as csvfile:
+            writer = csv.DictWriter(csvfile, fieldnames=SUMMARY_FIELDS)
+            writer.writeheader()
+
+    # Configure logging.
+    logging.basicConfig(format="{name}: {message}", style="{")
+    logger = logging.getLogger(parser.prog)
+
+    # Try current directory if pono is not on PATH.
+    executable = shutil.which("pono") or "./pono"
```
Po-Chun-Chien, line 137: Log error and abort if executable is not found.

scripts/parallel_pono.py
```diff
@@ -1,135 +1,186 @@
...
+                if process.returncode in SOLVED_RETURN_CODES:
+                    if process.stdout is None:
+                        logger.warning(f"{name} has no stdout")
+                        print(ReturnCode(process.returncode).name.lower())
+                    else:
+                        print(process.stdout.read())
+                    if args.witness_file:
+                        witness_name = witnesses[name]
+                        shutil.copy(witness_name, args.witness_file)
+                        os.remove(witness_name)
```
Po-Chun-Chien: I recommend storing witnesses as `pathlib.Path` objects.
This allows us to simplify the code in a more Pythonic way:

```suggestion
                        witnesses[name].rename(args.witness_file)
```

scripts/parallel_pono.py
```diff
@@ -1,135 +1,186 @@
...
+        runtime=f"{runtime:.1f}",
+        engine=engine,
+        command=" ".join(cmd),
+    )
+    with open(file, "a") as csvfile:
+        writer = csv.DictWriter(csvfile, fieldnames=SUMMARY_FIELDS)
+        writer.writerow(dataclasses.asdict(summary))
+
+
+def clean_up(processes: dict[str, subprocess.Popen[str]], verbose: bool):
```
Po-Chun-Chien: We also need to clean up the temp files stored in `witnesses`.

Suppose witnesses are `pathlib.Path` objects:

```python
for tmp_file in witnesses:
    tmp_file.unlink(missing_ok=True)
```

scripts/parallel_pono.py
```diff
@@ -1,135 +1,186 @@
...
+        for name, process in processes.items():
+            end_time = time.time()
+            if process.poll() is not None:  # process has finished
+                if args.summarize:
+                    runtime = end_time - start_times[name]
+                    cmd = cast(list[str], process.args)
+                    summarize(args.summarize, name, process.returncode, runtime, cmd)
+                if process.returncode in SOLVED_RETURN_CODES:
+                    if process.stdout is None:
+                        logger.warning(f"{name} has no stdout")
```
Po-Chun-Chien, line 174: (minor) FYI: It’s [recommended](https://docs.astral.sh/ruff/rules/logging-f-string/) to use %-style formatting rather than f-strings in Python logging.

CyanoKobalamyne: I know, but I don't like %-formatting. :P And this doesn't perform any computation, so it doesn't matter if it's eagerly formatted or not.

scripts/parallel_pono.py
CyanoKobalamyne: That will happen on line 159. You don't think it's enough?

scripts/parallel_pono.py
CyanoKobalamyne: Good point. Added a cleanup step.

scripts/parallel_pono.py
CyanoKobalamyne: Done.

scripts/parallel_pono.py
Po-Chun-Chien: Indeed. It should be sufficient.

On a related note, we could make the executable path a command-line option so it’s not restricted to `cwd` or `PATH`, giving users more flexibility. I’ll leave that decision up to you.

CyanoKobalamyne: I did think about that, but at least right now I don't feel that there is too much of a difference between executing
```
PATH="/some/folder/bin:$PATH" parallel_pono.py
```
and
```
parallel_pono.py --path "/some/folder/bin/pono"
```

CyanoKobalamyne: Although I guess we could allow specifying the name of the binary, in case it isn't `pono`. I could add that in a later PR.

PR #446, opened 2025-08-26
Fix witness export in shortest-CEX search for BMC

Po-Chun-Chien:
When the search range is 1, return early to preserve the solver state for witness extraction, before any further push/pop/assert operations.


Po-Chun-Chien:
This [line](https://github.com/stanford-centaur/pono/pull/446/files#diff-e323e7a00c9d827693de98ba3ceb0f2eabec6ffaebe8f4720d4929e3d4db3558R160) destroys the solver's `sat` state:

```c
      if (cex_upper_bound + 1 <= i) solver_->push();
```

CyanoKobalamyne (approved):
Thanks!

PR #451, opened 2025-08-28
k-induction: Fix error when solver returns unknown

CyanoKobalamyne:


engines/kinduction.cpp
```diff
@@ -356,6 +356,11 @@ bool KInduction::check_simple_path_lazy(int i)
     // we return immediately here and skip simple path checking.
     if (options_.kind_no_simple_path_check_) return false;
 
+    if (r.is_unknown()) {
+      kind_log_msg(2, "      ", "simple path check UNKNOWN");
+      return false;
+    }
```
Po-Chun-Chien: I would use `!r.is_sat()` to catch also `r.is_null()` (though it is unlikely to happen):

```suggestion
    if (!r.is_sat()) {
      kind_log_msg(2, "      ", "simple path check {}", r.to_string());
      return false;
    }
```

Po-Chun-Chien (approved):
Thanks! I did not find time to fix this earlier today.

PR #452, opened 2025-08-31
Don't error on getting unknown from solver

CyanoKobalamyne:
Bitwuzla (and other solvers) can return "unknown" when they can't compute an interpolant or encounter equality between constant arrays. We should handle this more gracefully by returning "unknown" instead of "error."

### DIFF
--- a/engines/interp_seq_mc.cpp
+++ b/engines/interp_seq_mc.cpp
@@ -22,7 +22,6 @@
 #include "smt-switch/utils.h"
 #include "smt/available_solvers.h"
 #include "utils/logger.h"
-#include "utils/term_analysis.h"
 
 using namespace smt;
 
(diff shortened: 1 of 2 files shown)


engines/interpolantmc.cpp
```diff
@@ -184,13 +184,15 @@ bool InterpolantMC::step(const int i)
...
+      if (r.is_unsat()) {
         throw PonoException("Internal error: Expecting satisfiable result");
+      } else {
+        throw InternalSolverException("solver could not compute concrete cex");
       }
       return false;
     } else if (r.is_unknown()) {
       // TODO: figure out if makes sense to increase bound and try again
-      throw PonoException("Interpolant generation failed.");
+      throw InternalSolverException("interpolant generation failed");
```
Po-Chun-Chien, line 195: We could provide more information:

```suggestion
      throw InternalSolverException("interpolant generation failed due to "
                                                              + r.get_explanation());
```

engines/interp_seq_mc.cpp
```diff
@@ -172,10 +171,14 @@ bool InterpSeqMC::step(int i)
       // replay it in the solver with model generation
       concrete_cex_ = true;
     } else if (r.is_sat() && !bmc_res.is_sat()) {
-      throw PonoException("Internal error: Expecting satisfiable result");
+      if (bmc_res.is_unsat()) {
+        throw PonoException("expecting satisfiable result");
+      } else {
+        throw InternalSolverException("solver could not compute BMC query");
```
Po-Chun-Chien, line 177: Minor rewording suggestion: To me "compute" sounds like constructing the formula rather than solving it.

```suggestion
        throw InternalSolverException("solver could not solve BMC query");
```

Po-Chun-Chien, line 177: Not sure about throwing `InternalSolverException`, since this would cross module boundaries and create tighter coupling to smt-switch. This exception class belongs to the smt-switch module. In Pono, exceptions should generally be `PonoException` or one of its derived classes.

PR #453, opened 2025-09-03
parallel porfolio: implement Python suggestions from ruff

CyanoKobalamyne:
* Convert paths from `str` to `pathlib.Path`
* Make boolean arguments keyword-only
* Annotate return types
* Use `tempfile.NamedTemporaryFile` as context manager instead of manually closing the file
* Quote cast argument
* Don't pre-format strings passed to `logging.Logger` methods
* Remove `else` from conditional where `if` has a return statement

### DIFF
--- a/scripts/parallel_pono.py
+++ b/scripts/parallel_pono.py
@@ -181,7 +181,7 @@ def main() -> int:
                 else:
                     print(process.stdout.read())
                 if args.witness_file:
-                    witnesses[name].rename(args.witness_file)
+                    shutil.move(witnesses[name], args.witness_file)
                 return process.returncode
 
     return ReturnCode.UNKNOWN.value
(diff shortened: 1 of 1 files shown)


scripts/parallel_pono.py
```diff
@@ -167,21 +175,20 @@ def main() -> int:
...
-                    clean_up({name: process}, witnesses, args.verbose)
+                    clean_up({name: process}, witnesses, verbose=args.verbose)
                     break
+                if process.stdout is None:
+                    logger.warning("%s has no stdout", name)
+                    print(ReturnCode(process.returncode).name.lower())
+                else:
+                    print(process.stdout.read())
+                if args.witness_file:
+                    witnesses[name].rename(args.witness_file)
```
Po-Chun-Chien: IIRC, we wanted to use `shutil.move` instead of `Path.rename` to ensure it works across filesystems.

CyanoKobalamyne: Oh yes, I made this PR before we added that fix. Updated now.

PR #456, opened 2025-09-10
Fix witness syntax for array value assignment

Po-Chun-Chien:
Btor2 witness format expects an array assignment contains a pair of index and value


Po-Chun-Chien:
Here's a small test case:

```
1 sort bitvec 1
2 sort bitvec 2
3 one 2
4 zero 2
5 sort array 2 2
6 state 5
7 init 5 6 3
8 next 5 6 6
9 input 5
10 read 2 6 4
11 read 2 9 4
12 add 2 10 11
13 eq 1 12 4
14 bad 13
```

PR #458, opened 2025-09-21
Allow users to pass options to SMT solver

Po-Chun-Chien:


options/options.h
```diff
@@ -372,12 +378,14 @@ class PonoOptions
   // TODO distinguish when solver is not set and choose a
   //      good solver for the provided engine automatically
   static const smt::SolverEnum default_smt_solver_ = smt::BZLA;
+  inline static const StringMap default_smt_solver_opts_ = {};
```
CyanoKobalamyne, line 381: Why do these have to be marked `inline`?

smt/msat_options.h
```diff
@@ -25,14 +25,21 @@
...
 
-using StringMap = std::unordered_map<std::string, std::string>;
-
 // configuration options copied from open-source ic3ia implementation
 // https://es-static.fbk.eu/people/griggio/ic3ia/index.html
 msat_config get_msat_config_for_ic3(bool interp, const StringMap & options)
 {
   msat_config cfg = msat_create_config();
 
+  for (const auto & optpair : options) {
```
CyanoKobalamyne: Why did you reorder these? I'm not sure how MathSAT behaves when an option is set twice, but if later ones just override previous ones, wouldn't it be better to allow users to override the defaults here?

options/options.cpp
```diff
@@ -201,20 +223,36 @@ const option::Descriptor usage[] = {
...
 #endif
   },
+  { SMT_INTERPOLATOR_OPT,
+    0,
+    "",
+    "smt-interpolator-opt",
+    Arg::ColonSepPair,
+    "  --smt-interpolator-opt <smt-opt>:<value> \tSet option in SMT "
+    "interpolator. May be specified multiple times. Does not overwrite "
+    "existing settings in Pono. (Expert option)" },
```
CyanoKobalamyne: Why do we not want users to be able to overwrite existing settings?

options/options.cpp
```diff
@@ -142,6 +144,26 @@ struct Arg : public option::Arg
       printError("Option '", option, "' requires a non-empty argument\n");
     return option::ARG_ILLEGAL;
   }
+
+  static option::ArgStatus ColonSepPair(const option::Option & option, bool msg)
+  {
+    if (option.arg != 0 && option.arg[0] != 0) {
```
CyanoKobalamyne: Shouldn't the first check here be `option.arg != nullptr` and the second `option.arg[0] != '\0'` (or, even better, `strlen(option.arg) > 0`), given that the type of `option.arg` is `const char *`?

options/options.h
Po-Chun-Chien: `inline` is needed for in-class initialization for non-literal type. Otherwise, we will need to add `const::PonoOptions::default_smt_solver_opts_ = {}` outside of class declaration.

options/options.cpp
Po-Chun-Chien: I’ve been thinking about this as well. There are certain options we don’t want users to overwrite, such as `produce-unsat-assumptions`, `incremental`, `produce-models`, and `produce-interpolants`, since changing them would break the engines.

How about we define a set of disallowed options that users cannot modify, and let them configure (and overwrite) the rest as they wish? Does that sound better to you?

smt/msat_options.h
Po-Chun-Chien: > if later ones just override previous ones

Yes, I think that’s the behavior, although it isn’t documented in their API.

options/options.cpp
CyanoKobalamyne: Yes, the way you have written it currently is definitely safer. We can leave it like this for now, I think. If we have a need to change these options from the command line later, we can revisit this.

options/options.cpp
Po-Chun-Chien: I simply copied the check from `NonEmpty`  above.
I'll apply your suggestions to both.

Po-Chun-Chien: Updated in 8deb6e2

options/options.cpp
Po-Chun-Chien: I decided to go with the proposal in https://github.com/stanford-centaur/pono/pull/458#discussion_r2376057168 (allowing users to overwrite solver options with some exceptions). This approach offers more flexibility and makes it easier, for example, to experiment with solver settings for IC3 using MathSAT.

CyanoKobalamyne (approved):

PR #461, opened 2025-09-23
Build pono with Boolector

CyanoKobalamyne:
Moved some code around so that when witness computation is not requested, Boolector works even without setting `base-context-level-1`.


smt/available_solvers.cpp
```diff
@@ -126,8 +126,12 @@ SmtSolver create_solver(SolverEnum se,
...
 }
 
-SmtSolver create_solver_for(
-    SolverEnum se, Engine e, bool logging, bool full_model, bool printing)
+SmtSolver create_solver_for(SolverEnum se,
+                            Engine e,
+                            bool logging,
+                            bool full_model,
+                            bool printing,
+                            bool witness)
```
Po-Chun-Chien: I’m considering whether we should pass a `PonoOptions` object here instead, as having so many boolean flags may be error-prone.

engines/bmc.cpp
```diff
@@ -45,13 +45,13 @@ void Bmc::initialize()
...
 
   logger.log(2, "BMC adding init constraint for step 0");
-  solver_->reset_assertions();
   solver_->assert_formula(unroller_.at_time(ts_.init(), 0));
 }
 
 void Bmc::reset_env()
 {
   initialized_ = false;
+  solver_->reset_assertions();
```
Po-Chun-Chien: I’d prefer putting `reset_assertions()` in `initialize()`, since the solver may also be used externally for other checks. This ensures that the first call to `check_until()` always starts from a clean solver state.

smt/available_solvers.cpp
```diff
@@ -174,6 +178,10 @@ SmtSolver create_solver_for(
   assert(s);
   if (ic3_engine) {
     s->set_opt("produce-unsat-assumptions", "true");
+    if (se == BTOR && witness) {
+      // Witness generation in IC3 needs reset_assertions support.
+      s->set_opt("base-context-1", "true");
+    }
```
Po-Chun-Chien: Not only witness generation but also some engines rely on `reset_assertions()`. I think in Pono we should simply enable this by default for Boolector.

While this may impact performance, it avoids adding more logic for special cases and makes the code easier to maintain. Since Boolector is no longer our primary solver, I think this tradeoff is acceptable.

smt/available_solvers.cpp
Po-Chun-Chien: If we decide to not add `bool witness` (as proposed in https://github.com/stanford-centaur/pono/pull/461#discussion_r2375946204), then we can leave it as is for now.

CyanoKobalamyne: I was thinking that we could fold these into the options map once #458 is merged.

CyanoKobalamyne: Ended up reverting this.

engines/bmc.cpp
CyanoKobalamyne: Reverted this.

smt/available_solvers.cpp
CyanoKobalamyne: Good point.

PR #463, opened 2025-09-23
parallel portfolio: Allow specifying pono binary name

CyanoKobalamyne:


scripts/parallel_pono.py
```diff
@@ -113,10 +113,24 @@ def clean_up(
...
+    """Return the path to a file in PATH or the script's directory."""
+    fullname = shutil.which(name)
+    if fullname is None:
+        path = pathlib.Path(__file__).parent / name
+    else:
+        path = pathlib.Path(fullname)
+    if not path.is_file():
+        msg = f"file {name} not in script directory or PATH"
+        raise FileNotFoundError(msg)
+    return path
```
Po-Chun-Chien, line 137: We could also check whether `path` is executable (in case it’s not found by `which`):

```suggestion
    if not os.access(path, os.X_OK):
        raise PermissionError(f"{path} is not executable")
    return path
```

scripts/parallel_pono.py
```diff
@@ -132,15 +146,13 @@ def main() -> int:
...
-    # Try current directory if pono is not on PATH.
-    executable = shutil.which("pono") or pathlib.Path(__file__).parent / "pono"
-
     processes: dict[str, subprocess.Popen[str]] = {}
     start_times: dict[str, float] = {}
     witnesses: dict[str, pathlib.Path] = {}
     atexit.register(clean_up, processes, witnesses, args.verbose)
 
     # Launch each portfolio solver as a subprocess.
+    executable = find_file(args.binary)
```
Po-Chun-Chien: (minor nitpick) `find_executable` might be a clearer name for the method

scripts/parallel_pono.py
```diff
@@ -113,10 +113,24 @@ def clean_up(
                 logger.warning(line)
 
 
+def find_file(name: str) -> pathlib.Path:
+    """Return the path to a file in PATH or the script's directory."""
```
Po-Chun-Chien, line 125: In addition to `PATH` and script dir, we can also try to find the binary in `cwd`.

scripts/parallel_pono.py
```diff
@@ -113,10 +113,24 @@ def clean_up(
...
+        msg = f"file {name} not in script directory or PATH"
+        raise FileNotFoundError(msg)
+    return path
+
+
 def main() -> int:
     parser = argparse.ArgumentParser(description="Run multiple engines in parallel")
     parser.add_argument("btor_file", help="input benchmark in BTOR2 format")
     parser.add_argument("witness_file", nargs="?", help="file to store the witness")
+    parser.add_argument("-b", "--binary", default="pono", help="name of pono binary")
```
Po-Chun-Chien: (minor) Add a bit more info for the user:

```suggestion
    parser.add_argument("-b", "--binary", default="pono", help="name of pono binary (search in PATH, script dir, and CWD)")
```

scripts/parallel_pono.py
CyanoKobalamyne: For now I would prefer to keep it this way. I think it's good if the script behaves the same, regardless from where it is invoked. If it checks the current working directory, it could inadvertently pick up an unintended version if the user is not paying attention to what directory they are in.

scripts/parallel_pono.py
CyanoKobalamyne: Good idea.

scripts/parallel_pono.py
CyanoKobalamyne: Good point.

PR #465, opened 2025-09-23
parallel portfolio: Better help message

CyanoKobalamyne:


scripts/parallel_pono.py
```diff
@@ -114,12 +114,20 @@ def clean_up(
...
-    parser = argparse.ArgumentParser(description="Run multiple engines in parallel")
+    parser = argparse.ArgumentParser(
+        description="Run multiple engines in parallel",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
     parser.add_argument("btor_file", help="input benchmark in BTOR2 format")
-    parser.add_argument("witness_file", nargs="?", help="file to store the witness")
-    parser.add_argument("-k", "--bound", default=1000, type=int, help="unrolling bound")
+    parser.add_argument("witness_file", nargs="?", help="file to store the witness in")
+    parser.add_argument("-k", "--bound", default=1000, type=int, help="check bound")
```
Po-Chun-Chien: "check bound" seems a bit unclear to me. We can use the same help message as in Pono.

```suggestion
    parser.add_argument("-k", "--bound", default=1000, type=int, help="bound to check up until")
```

PR #467, opened 2025-09-24
Update MathSAT to 5.6.12

CyanoKobalamyne:


CyanoKobalamyne:
This actually makes pono's solving performance with IC3IA on the HWMCC array benchmarks slightly worse, so let's not merge it for now.

PR #469, opened 2025-09-26
Fix base case when `--kind-one-time-base-check` is enabled

### LINKED ISSUE #457: False proof caused by `--kind-one-time-base-check`
Pono incorrectly returned `unsat` on this task: [simple_4-1_abstracted.btor2.gz](https://github.com/user-attachments/files/22428620/simple_4-1_abstracted.btor2.gz)

- Pono version: 74fddc7ef7d177a91b657eb2d2fd91e22b32e610
- Command:

  ```bash
  ./pono -e ind --kind-one-time-base-check simple_4-1_abstracted.btor2
  ```

Without `--kind-one-time-base-check`, Pono correctly returned `sat`.

**TLDR**: `--kind-one-time-base-check` could cause unsoundness if the transition relation is not right-total.

### More details

If `--kind-one-time-base-check` is enabled, Pono checks the inductive case first, and only proceeds to the base case if the former is `unsat`.

- Inductive case: $`P(s_0) \wedge T(s_0, s_1) \ldots P(s_{k-1}) \wedge T(s_{k-1}, s_k) \wedge \neg P(s_k)`$
- Base case: $`Init(s_0) \wedge T(s_0, s_1) \ldots T(s_{k-1}, s_k) \wedge (\neg P(s_0) \vee \neg P(s_k))`$

This procedure can be unsound if $`T(s, s')`$ is not right-total, that is, $`Init(s_0) \wedge T(s_0, s_1) \ldots T(s_{k-1}, s_k)`$ or $`T(s_0, s_1) \ldots T(s_{k-1}, s_k)`$ is already `unsat` by itself.

For the above instance `simple_4-1_abstracted.btor2`:

$`T(s_0, s_1) \wedge T(s_1, s_2)`$ is `unsat`. Therefore, both inductive and base cases at `k = 2` are `unsat`.
However, the base case at `k = 1` is sat.

Po-Chun-Chien:
This PR fixes #457 and introduces a precise check for the right-totalness of the transition relation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ RUN apt-get install -y \
         curl \
         git \
         libgmp-dev \
+        libbison-dev \
         m4 \
         meson \
         ninja-build \
@@ -22,7 +23,6 @@ RUN apt-get install -y \
         wget
 RUN apt-get autoremove --purge
 RUN apt-get clean -y
-RUN ./contrib/setup-bison.sh
 RUN ./contrib/setup-flex.sh
 RUN ./contrib/setup-btor2tools.sh
 


Po-Chun-Chien:
We should also add some engine tests on transition systems with non-total transition relations, but I don't have time ATM.

engines/kinduction.cpp
```diff
@@ -425,14 +459,33 @@ bool KInduction::final_base_case_check(int cur_bound)
...
+  solver_->assert_formula(init0_);
+  Term query = false_;
+  for (int i = cur_bound; i >= 0; --i) {
+    Term bad_i = unroller_.at_time(bad_, i);
+    query = solver_->make_term(PrimOp::Or, bad_i, query);
+    if (i > 0) {
+      Term tr_to_i = unroller_.at_time(ts_.trans(), i - 1);
+      query = solver_->make_term(PrimOp::And, tr_to_i, query);
+    }
+  }
```
CyanoKobalamyne: This could be slightly simplified to:
```suggestion
  Term query = unroller_.at_time(bad_, 0);
  for (int i = cur_bound; i >= 1; --i) {
    Term bad_i = unroller_.at_time(bad_, i);
    query = solver_->make_term(PrimOp::Or, bad_i, query);
    Term tr_to_i = unroller_.at_time(ts_.trans(), i - 1);
    query = solver_->make_term(PrimOp::And, tr_to_i, query);
  }
```

CyanoKobalamyne: Also, is there a reason why you used decrement here instead of increment? It doesn't really matter, just curious.

engines/kinduction.cpp
```diff
@@ -402,9 +402,43 @@ void KInduction::kind_log_msg(size_t level,
...
+  const bool is_trans_total = [](const PonoOptions & opt,
+                                 const TransitionSystem & ts) -> bool {
+    try {
+      return (opt.check_trans_total_ && ts.is_right_total())
+             || (ts.is_functional() && ts.constraints().empty());
+    }
+    catch (SmtException &) {
+      return false;
+    }
+  }(options_, ts_);
```
CyanoKobalamyne: I think you should put this in a separate helper function `is_trans_total`, because this is a little difficult to parse.

engines/kinduction.cpp
```diff
@@ -402,9 +402,43 @@ void KInduction::kind_log_msg(size_t level,
...
+
+  if (is_trans_total) {
+    prep_final_base_case_check_total(cur_bound);
+  } else {
+    prep_final_base_case_check_nontotal(cur_bound);
+  }
```
CyanoKobalamyne: I think you can just inline these. If you move the stuff above into a separate function, this will still be a manageable size. The `prep` functions don't make much sense independently.

CyanoKobalamyne: Also, the first (`sel_assumption_.clear()`) and last (`solver_->assert_formula(x)`) statements are actually the same, so there is room to eliminate some redundancy here.

core/ts.h
```diff
@@ -300,6 +300,35 @@ class TransitionSystem
...
+   * If the system is functional and has no constraints, it is right-total.
+   *
+   * Otherwise, a fresh solver instance is created and checks if the following
+   * query is unsatisfiable.
+   *
+   * - If `inputs_as_states` is false (default), the query is:
+   *   `exists curr. forall next. not (exist input. trans(curr, input, next))`
+   *
+   * - If `inputs_as_states` is true, the query is:
+   *   `exist {curr, input}. forall next. not trans(curr, input, next)`.
```
CyanoKobalamyne: What is the scenario where `inputs_as_states` would be set to true? Real inputs that for some values cause the transition relation to not be right total are fine, since we will still find the SAT solution where they are set to a value where there is indeed a next state. It is only when the relations doesn't hold for any current state and *any* input that is the problem.

core/ts.cpp
```diff
@@ -629,6 +631,100 @@ bool TransitionSystem::contains(const Term & term,
...
+      throw e;
+    }
+    logger.log(
+        1,
+        "WARNING: Right-total check using {} failed, trying again with {}",
+        to_string(se),
+        to_string(fallback_se));
+    return is_right_total(fallback_se, inputs_as_states);
+  }
+}
```
CyanoKobalamyne: I think if the solver claims to support quantifiers but can't solve the query, it's fine to just throw the error; I don't think we need all this logic for falling back to cvc5. If the solver does not support quantifiers, then we can just use cvc5. I guess we can still exclude Boolector if needed. Did you get segfaults when testing this?

CyanoKobalamyne: Also, perhaps we should move this stuff into `smt/available_solvers.h` as `create_quantifier_solver()` or something like that.

core/ts.cpp
```diff
@@ -629,6 +631,100 @@ bool TransitionSystem::contains(const Term & term,
...
+    query = s->make_term(Exists, input_params);
+  }
+  next_params.push_back(s->make_term(Not, query));
+  query = s->make_term(Forall, next_params);
+  if (inputs_as_states) {
+    input_params.push_back(query);
+    query = s->make_term(Exists, input_params);
+  }
+  curr_params.push_back(query);
+  query = s->make_term(Exists, curr_params);
```
CyanoKobalamyne: Can't we omit the outer existential quantifiers here? Since the free variables will just be implicitly existential, no?

core/ts.cpp
```diff
@@ -629,6 +631,100 @@ bool TransitionSystem::contains(const Term & term,
...
+    next_params.push_back(p);
+  }
+
+  // construct query
+  // if unsat, the transition relation is right-total
+  Term query = s->substitute(tt.transfer_term(trans(), BOOL), var_map);
+  if (!inputs_as_states) {
+    input_params.push_back(query);
+    query = s->make_term(Exists, input_params);
+  }
```
CyanoKobalamyne: Same comment as for `ts.h`. When is this case used/useful?

pono.cpp
```diff
@@ -61,6 +61,9 @@ ProverResult check_prop(PonoOptions pono_options,
              to_string(pono_options.smt_interpolator_));
   logger.log(3, "INIT:\n{}", ts.init());
   logger.log(3, "TRANS:\n{}", ts.trans());
+  if (pono_options.check_trans_total_) {
+    logger.log(1, "TRANS is {}right-total", ts.is_right_total() ? "" : "NOT ");
+  }
```
CyanoKobalamyne: Maybe move this logging statement to the other call site where we call `ts.is_right_total()` in`engines/kinduction.cpp`, because we will currently make two identical queries when this option is turned on.

engines/kinduction.cpp
Po-Chun-Chien: See d646d56

engines/kinduction.cpp
Po-Chun-Chien: See e70bfca

pono.cpp
Po-Chun-Chien: See 8ffecc0

core/ts.h
Po-Chun-Chien: See c806306

core/ts.cpp
Po-Chun-Chien: See b3c20a5.

> Did you get segfaults when testing this?

Yes, Boolector ran into segfaults (both in unit test and the Btor2 instance in #457). Similarly, MathSAT returned `unknown`.

engines/kinduction.cpp
Po-Chun-Chien: The procedure you proposed will lead to a different (and incorrect) query.
Suppose `cur_bound` is 1.

- Original: `Bad(0) || (Tr(0, 1) && bad(1))`
- Proposed: `Tr(0, 1) && (Bad(0) || Bad(1))`

The latter still has the same issue as in #457.
If `Tr(0, 1)` or `Init(0) && Tr(0, 1)` is unsat, it will mask the cex in `Init(0) && Bad(0)`.

> Also, is there a reason why you used decrement here instead of increment? It doesn't really matter, just curious.

The order **does** matter. `Bad(cur_bound)` has to be created first as it is the innermost term in the nested formula (note the alternation of `And` and `Or` in the loop).

core/ts.cpp
Po-Chun-Chien: I am not sure if all solvers behave this way.

I tested with cvc5. It returned the same answer with or without the outermost existential quantifier.
However, Bitwuzla, crashed (segfault) when the quantifier was omitted.

In any case, I think it's safer to ensure the correct semantic from the Pono side.

Po-Chun-Chien: (BTW, could you re-enable Bitwuzla's interpolation support in smt-switch https://github.com/stanford-centaur/smt-switch/pull/429 ?)

engines/kinduction.cpp
```diff
@@ -397,42 +397,84 @@ template <typename... Args>
...
+    sel_assumption_.push_back(sel_simple_path_terms_);
+
+    // build a disjunctive bad state property ranging over bounds 0,...,i
+    // Potential optimization(?): do we have to include the bad state property
+    // for bound 'i' or can we omit it as the respective BMC problem for bound
+    // 'i' is unsat when the inductive check is unsat at bound 'i'?
+    for (int frame = 0; frame <= cur_bound; frame++) {
+      query =
+          solver_->make_term(PrimOp::Or, query, unroller_.at_time(bad_, frame));
+    }
```
CyanoKobalamyne:
```suggestion
    Term query = unroller_.at_time(bad_, 0)
    for (int frame = 1; frame <= cur_bound; frame++) {
      query =
          solver_->make_term(PrimOp::Or, query, unroller_.at_time(bad_, frame));
    }
```

engines/kinduction.cpp
CyanoKobalamyne: Oh okay, yeah, I misread this code.

engines/kinduction.cpp
```diff
@@ -397,42 +397,84 @@ template <typename... Args>
...
-  // 'i' is unsat when the inductive check is unsat at bound 'i'?
-  Term disj = false_;
-  int frame;
-  for (frame = 0; frame <= cur_bound; frame++) {
-    disj = solver_->make_term(PrimOp::Or, disj, unroller_.at_time(bad_, frame));
-  }
-  solver_->assert_formula(disj);
   kind_log_msg(
       1, "", "checking base case a posteriori at bound: {}", cur_bound);
+  Term query = false_;
```
CyanoKobalamyne:
```suggestion
```

engines/kinduction.cpp
```diff
@@ -397,42 +397,84 @@ template <typename... Args>
...
+    solver_->assert_formula(init0_);
+    for (int i = cur_bound; i >= 0; --i) {
+      Term bad_i = unroller_.at_time(bad_, i);
+      query = solver_->make_term(PrimOp::Or, bad_i, query);
+      if (i > 0) {
+        Term tr_to_i = unroller_.at_time(ts_.trans(), i - 1);
+        query = solver_->make_term(PrimOp::And, tr_to_i, query);
+      }
+    }
+  }
```
CyanoKobalamyne:
```suggestion
    Term query = unroller_.at_time(bad_, cur_bound);
    for (int i = cur_bound - 1; i >= 0; --i) {
        Term tr_from_i = unroller_.at_time(ts_.trans(), i);
        query = solver_->make_term(PrimOp::And, tr_from_i, query);
        Term bad_i = unroller_.at_time(bad_, i);
        query = solver_->make_term(PrimOp::Or, bad_i, query);
    }
```

CyanoKobalamyne (commented):
I have reworked the suggestions for simplifying the loops, so I think they are correct now.

core/ts.cpp
```diff
@@ -629,6 +630,69 @@ bool TransitionSystem::contains(const Term & term,
   return true;
 }
 
+bool TransitionSystem::is_right_total() const
```
ahmed-irfan, line 633: maybe store the result in a variable and a variable to whether check again (if more constraints are added to trans) or not.

Po-Chun-Chien: I’ve considered that as well. The downside is that it would make `is_right_total(`) non-const, since it would modify internal members of TS. We’d also need to maintain these new class members across other API calls. At the moment, it’s only used once in the k-induction engine, so repeated calls aren’t an issue. If we later find that multiple repeated calls are needed, we can revisit this approach.

engines/kinduction.cpp
Po-Chun-Chien: Applied in d9ea343 along with the other 2 suggestions and some small tweaks.

Po-Chun-Chien:
The CI is failing because the TS in unit test has no input and Bitwuzla fails when quantifying an empty set of variable. I will fix it later.

PR #470, opened 2025-09-26
Soundly add monitor state

### LINKED ISSUE #380: Unsoundness due to property monitor
This issue follows up on #377.

Pono introduces a property-monitor state when the property depends on inputs.
The next-state value of the monitor reflects the property's evaluation at the current time frame.

However, this transformation can be unsound if the transition relation is not right-total.
Specifically, if a bad state in the original transition system has no successor,
the corresponding counterexample may be masked in the transformed circuit.

This issue has been resolved for Btor2 models (#379),
but it still persists for other models, such as VMT and SMV.

Po-Chun-Chien:
Fixes #380

### DIFF
--- a/modifiers/prop_monitor.h
+++ b/modifiers/prop_monitor.h
@@ -61,14 +61,19 @@ smt::Term add_prop_monitor(TransitionSystem & ts, const smt::Term & prop)
   smt::SmtSolver s = ts.solver();
   // monitor is set to false if prop is ever false:
   // monitor' = prop && monitor
-  if (ts.no_next(prop)) {
-    ts.assign_next(monitor, s->make_term(smt::And, prop, monitor));
-  } else {
-    assert(!ts.is_functional());
+  smt::Term monitor_next = s->make_term(smt::And, prop, monitor);
+  if (ts.is_functional()) {
+    assert(ts.no_next(prop));
+    ts.assign_next(monitor, monitor_next);
+  } else {  // relational TS
     RelationalTransitionSystem & rts =
         static_cast<RelationalTransitionSystem &>(ts);
-    rts.constrain_trans(rts.make_term(
-        smt::Equal, rts.next(monitor), s->make_term(smt::And, prop, monitor)));
+    if (rts.no_next(prop)) {
+      rts.assign_next(monitor, monitor_next);
+    } else {
+      rts.constrain_trans(
+          rts.make_term(smt::Equal, rts.next(monitor), monitor_next));
+    }
     // ensure that if prop is false, there will always be a next state
     rts.set_trans(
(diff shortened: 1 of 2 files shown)


modifiers/prop_monitor.h
```diff
@@ -22,10 +22,34 @@
...
+  // only in debug mode as right-total check can be expensive
+  assert(!ts.no_next(prop) || ts.is_right_total());
   logger.log(1, "Adding a monitor for the property");
 
+  // checks for functional TS
+  if (ts.is_functional() && !ts.no_next(prop)) {
+    throw PonoException(
+        "Cannot use next in property of a functional transition system.");
+  }
+
```
CyanoKobalamyne: What's the reason for moving this call here? Now this function will call `ts.no_next()` twice, which adds some overhead, given that it has to traverse the property twice.

CyanoKobalamyne (commented):
Thanks, doing this refactoring was on my to-do list for some time.

modifiers/prop_monitor.h
Po-Chun-Chien: I think it’s better to perform the precondition check before modifying the transition system (e.g., adding a new state). I refactored this in 9be30eb. Now `ts.no_next()` is called only once.

modifiers/prop_monitor.cpp
```diff
@@ -0,0 +1,67 @@
+/*********************                                                  */
+/*! \file prop_monitor.cpp
+** \verbatim
+** Top contributors (to current version):
+**   Makai Mann, Ahmed Irfan
```
ahmed-irfan: add your name here

modifiers/prop_monitor.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "core/ts.h"
+#include "utils/logger.h"
+
+namespace pono {
+
+smt::Term add_prop_monitor(TransitionSystem & ts, const smt::Term & prop)
+{
+  const bool prop_has_next = !ts.no_next(prop);
+  // only in debug mode as right-total check can be expensive
+  assert(prop_has_next || ts.is_right_total());
```
ahmed-irfan, line 29: if it is not right total, what will happen in the release mode?

modifiers/prop_monitor.cpp
Po-Chun-Chien: Done in e800c84

modifiers/prop_monitor.cpp
Po-Chun-Chien: Then the monitor will still be added.

However, with #476, `add_prop_monitor` will only be called when `prop_has_next` is `true`.

PR #474, opened 2025-09-28
Fix Relational Transition Systems soundness bugs

augustomafra:
Since Relational Transition Systems might have their state transitions set without calling TransitionSystem::assign_next, valid states would be incorrectly marked as not having any transitions (by remaining into the "TransitionSystem::no_state_updates_" set).

This caused incorrect proven results on BMC-SP and k-induction engines. Fixed by traversing the transition constrain in RTS and ensuring that all state variables are removed from "no_state_updates_".

I'm adding test cases covering this fix in a separate pull request because the SMV inputs reproducing the bug depend on other SMV frontend enhancements.

### DIFF
--- a/core/rts.cpp
+++ b/core/rts.cpp
@@ -59,7 +59,7 @@ void RelationalTransitionSystem::set_updated_states(const smt::Term & term)
   if (curr != curr_map_.end()) {
     no_state_updates_.erase(curr->second);
   }
-  for (const auto& subTerm : *term) {
+  for (const auto & subTerm : *term) {
     set_updated_states(subTerm);
   }
 }
--- a/core/rts.cpp
+++ b/core/rts.cpp
@@ -30,6 +30,7 @@ void RelationalTransitionSystem::set_behavior(const Term & init,
   }
   init_ = init;
   trans_ = trans;
+  set_updated_states(trans_);
 }
 
 void RelationalTransitionSystem::set_trans(const Term & trans)
@@ -39,6 +40,7 @@ void RelationalTransitionSystem::set_trans(const Term & trans)
     throw PonoException("Unknown symbols");
   }
   trans_ = trans;
+  set_updated_states(trans_);
 }
 
 void RelationalTransitionSystem::constrain_trans(const Term & constraint)
@@ -48,6 +50,17 @@ void RelationalTransitionSystem::constrain_trans(const Term & constraint)
     throw PonoException("Unknown symbols");
   }
   trans_ = solver_->make_term(And, trans_, constraint);
+  set_updated_states(constraint);
 }
 
+void RelationalTransitionSystem::set_updated_states(const smt::Term & term)
+{
+  auto curr = curr_map_.find(term);
+  if (curr != curr_map_.end()) {
+    no_state_updates_.erase(curr->second);
+  }
+  for (const auto& subTerm : *term) {
+    set_updated_states(subTerm);
+  }
+}
 }  // namespace pono
--- a/engines/kinduction.cpp
+++ b/engines/kinduction.cpp
@@ -293,7 +293,9 @@ Term KInduction::simple_path_constraint(int i, int j)
     disj = solver_->make_term(PrimOp::Or, disj, neq);
   }
   // add selector term
-  disj = solver_->make_term(PrimOp::Or, disj, sel_simple_path_terms_);
+  if (disj != false_) {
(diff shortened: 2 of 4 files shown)


Po-Chun-Chien (approved):
LGTM. Suggested a small improvement.

core/rts.cpp
```diff
@@ -48,6 +50,17 @@ void RelationalTransitionSystem::constrain_trans(const Term & constraint)
...
 
+void RelationalTransitionSystem::set_updated_states(const smt::Term & term)
+{
+  auto curr = curr_map_.find(term);
+  if (curr != curr_map_.end()) {
+    no_state_updates_.erase(curr->second);
+  }
+  for (const auto & subTerm : *term) {
+    set_updated_states(subTerm);
+  }
```
Po-Chun-Chien: Consider using `get_free_symbolic_consts` (`smt-switch/utils.h`) to extract inputs and state variables from the term, and filter for those in `next_statevars_`.

engines/kinduction.cpp
Po-Chun-Chien, line 1: Why are the changes here necessary? Wouldn’t the fix in `rts` be sufficient?

My understanding is that if the system has no states with update (i.e., it is purely combinational), the simple-path constraint is trivially unsat.

And AFAICS, the simple-path constraints are only added when `t > 0`.
So if a property violation occurs at `t = 0`, it should still be detected.

Po-Chun-Chien (changes requested):

engines/kinduction.cpp
augustomafra: Sounds good, I'm removing it on f47a89e08bac597b9ff29de94d1c4676ad22f612.
This was my first fix, but it is subsumed by the solution on rts.cpp. I left it here conservatively in case it could catch any other scenario I missed. But it does indeed look redundant.

core/rts.cpp
augustomafra: Thanks for the note, updated at 12c71bcb5d6420b8f5a83e65c296ff0ffc1a9786

core/rts.cpp
```diff
@@ -48,6 +52,17 @@ void RelationalTransitionSystem::constrain_trans(const Term & constraint)
...
 }
 
+void RelationalTransitionSystem::set_updated_states(const smt::Term & term)
+{
+  UnorderedTermSet free_vars;
+  get_free_symbolic_consts(term, free_vars);
+  for (const auto & free_var : free_vars) {
+    if (next_statevars_.count(free_var)) {
+      no_state_updates_.erase(free_var);
+    }
```
Po-Chun-Chien: There seems to be a mistake. The next-state var has to mapped back to curr-state var before erasing.
Could you double-check?

```suggestion
    if (next_statevars_.find(free_var) != next_statevars_.end()) {
      no_state_updates_.erase(curr_map_[free_var]);
    }
```

Po-Chun-Chien (changes requested):
Could you also add a unit test in `tests/test_ts.cpp`:

```c
TEST_P(TSUnitTests, RTS_ConstrainTrans)
{
  RelationalTransitionSystem rts(s);
  Term x = rts.make_statevar("x", bvsort);
  Term next_x = rts.next(x);
  rts.constrain_trans(rts.make_term(BVUge, x, next_x));
  EXPECT_TRUE(rts.statevars_with_no_update().empty());
}
```

core/rts.cpp
augustomafra: You are correct, I was missing getting from the curr_map. Thanks for the thorough review.
Fixed at a9fa4b597cc5fdc2fe14f44a4fcbd84308e7c8ef.
Also adding the unit test and some of the original tests from #475 as they are required for covering this fix.

Po-Chun-Chien (approved):

core/rts.cpp
```diff
@@ -30,6 +32,7 @@ void RelationalTransitionSystem::set_behavior(const Term & init,
   }
   init_ = init;
   trans_ = trans;
+  set_updated_states(trans_);
```
CyanoKobalamyne: You could also just remove this call as well as the assignment and check above and just call `set_trans`.

CyanoKobalamyne (approved):

ahmed-irfan (approved):

core/rts.cpp
augustomafra: Sounds good, factored out that call at 5bfc28391c3fc71f8ddf858a8e58e0475a925810

PR #475, opened 2025-09-28
Support SMV resize and type-conversion operators

augustomafra:
SMV files generated from Yosys use several bit-vector resize and type conversion operations that are not currently supported by pono.

Also adding test cases for soundness bug on RTS that got exposed when testing Yosys-generated SMV files (see soundness fix PR at  #474 ).


frontends/smv_node.cpp
```diff
@@ -829,10 +829,11 @@ void pono::resize_expr::generate_ostream(
     std::unordered_map<string, string> new_prefix,
     ostream & s)
 {
-  s << " resize ";
+  s << " resize (";
   ex1->generate_ostream(name, prefix, module_list, new_prefix, s);
   s << " , ";
   ex2->generate_ostream(name, prefix, module_list, new_prefix, s);
+  s << " ) ";
```
CyanoKobalamyne, line 836: What is the purpose of this change?

frontends/smvparser.y
```diff
@@ -74,7 +74,7 @@ DOT ".";
 
 %right OP_IMPLY
 %left OP_BI
-%left IF_ELSE
+%right IF_ELSE
```
CyanoKobalamyne, line 77: I can't find the associativity of the ternary conditional described anywhere in the NuSMV manual; why did you make this change?

frontends/smvparser.y
```diff
@@ -852,7 +852,15 @@ simple_expr: constant {
...
-                e = enc.solver_->make_term(smt::Equal, a->getTerm(), b->getTerm());
+                smt::Term a_term = a->getTerm();
+                smt::Term b_term = b->getTerm();
+                if (bvs_a == SMVnode::Real || bvs_b == SMVnode::Real){
+                  if (bvs_a != SMVnode::Real)
+                    a_term = enc.solver_->make_term(smt::To_Real, a_term);
+                  if (bvs_b != SMVnode::Real)
+                    b_term = enc.solver_->make_term(smt::To_Real, b_term);
+                }
+                e = enc.solver_->make_term(smt::Equal, a_term, b_term);
```
CyanoKobalamyne, line 863: As per the NuSMV manual, mixed-type equality check are not actually allowed.

frontends/smvparser.y
```diff
@@ -1202,13 +1210,51 @@ simple_expr: constant {
               }
             }
             | word1 "(" basic_expr ")" {
-              throw PonoException("No word1");
+              if(enc.module_flat){
+                SMVnode *boolean = $3;
+                if(boolean->getType() != SMVnode::Boolean){
+                  throw PonoException("Word1 word type is uncompatible");
```
CyanoKobalamyne:
```suggestion
                  throw PonoException("word1 can only be applied to booleans");
```

frontends/smvparser.y
```diff
@@ -1202,13 +1210,51 @@ simple_expr: constant {
...
+                }
+                smt::Sort bv1sort = enc.solver_->make_sort(smt::BV, 1);
+                smt::Term res = enc.solver_->make_term(smt::Ite, 
+                                                       boolean->getTerm(), 
+                                                       enc.solver_->make_term(1, bv1sort), 
+                                                       enc.solver_->make_term(0, bv1sort));
+                assert(res); //check res non-null
```
CyanoKobalamyne: I don't think this is necessary. When would `res` be null? In any case, assertions get compiled out of non-debug builds.

frontends/smvparser.y
```diff
@@ -1202,13 +1210,51 @@ simple_expr: constant {
...
+              }else{
+                $$ = new word1_expr($3);
+              }
             }
             | tok_bool "(" basic_expr ")"{
-              throw PonoException("No type convert");
+              if(enc.module_flat){
+                SMVnode *a = $3;
+                smt::Sort sort = a->getTerm()->get_sort();
+                // TODO: SMV also supports bool conversion from integer
```
CyanoKobalamyne: Please add the integer conversion too; it should require little code.

frontends/smvparser.y
```diff
@@ -1202,13 +1210,51 @@ simple_expr: constant {
...
+              }
             }
             | tok_toint "(" basic_expr ")"{
-              throw PonoException("No type convert");
+              if(enc.module_flat){
+                SMVnode *a = $3;
+                smt::Sort sort = a->getTerm()->get_sort();
+                if(sort->get_sort_kind() != smt::BV){
+                  throw PonoException("Can't convert non-bitvector to integer.");
+                }
```
CyanoKobalamyne: Booleans can also be converted to integer.

frontends/smvparser.y
```diff
@@ -1253,11 +1299,63 @@ simple_expr: constant {
...
               throw PonoException("No extend now");
             }
-            | resize "(" basic_expr ")" {
-              throw PonoException("No resize");
+            | resize "(" basic_expr "," integer_val ")" {
+              if(enc.module_flat){
+                SMVnode *word = $3;
+                SMVnode::Type word_type = word->getType();
+                
+                if(word_type != SMVnode::Integer && word_type != SMVnode::Signed && word_type != SMVnode::Unsigned){
```
CyanoKobalamyne: Resize is only allowed for bit-vectors.

frontends/smvparser.y
```diff
@@ -1253,11 +1299,63 @@ simple_expr: constant {
...
+                  throw PonoException("Resize word type is uncompatible");
+                }
+                
+                int integer = stoi($5);
+                smt::Sort word_sort = word->getTerm()->get_sort();
+                uint64_t word_width = word_type != SMVnode::Integer ? word_sort->get_width() : 0;
+
+                if (word_type == SMVnode::Integer){
+                  smt::Term res = enc.solver_->make_term(smt::Op(smt::Int_To_BV, integer), word->getTerm());
+                  assert(res);
```
CyanoKobalamyne: Same remark as for assert above.

frontends/smvparser.y
```diff
@@ -1253,11 +1299,63 @@ simple_expr: constant {
...
+                }else if(integer < word_width){
+                  smt::Term res;
+                  if(word_type == SMVnode::Signed){
+                    smt::Term tail = enc.solver_->make_term(smt::Op(smt::Extract, integer - 2, 0), word->getTerm());
+                    smt::Term signBit = enc.solver_->make_term(smt::Op(smt::Extract, word_width - 1, word_width - 1), word->getTerm());
+                    res = enc.solver_->make_term(smt::Concat, signBit, tail);
+                  }else{
+                    res = enc.solver_->make_term(smt::Op(smt::Extract, integer - 1, 0), word->getTerm());
+                  }
+                  assert(res); //check res non-null
```
CyanoKobalamyne: Same remark as for assert above.

frontends/smvparser.y
```diff
@@ -1253,11 +1299,63 @@ simple_expr: constant {
...
+            | resize "(" basic_expr "," integer_val ")" {
+              if(enc.module_flat){
+                SMVnode *word = $3;
+                SMVnode::Type word_type = word->getType();
+                
+                if(word_type != SMVnode::Integer && word_type != SMVnode::Signed && word_type != SMVnode::Unsigned){
+                  throw PonoException("Resize word type is uncompatible");
+                }
+                
+                int integer = stoi($5);
```
CyanoKobalamyne: This should be called something like `to_width` instead of `integer`.

frontends/smvparser.y
```diff
@@ -1253,11 +1299,63 @@ simple_expr: constant {
...
+                    res = enc.solver_->make_term(smt::Concat, signBit, tail);
+                  }else{
+                    res = enc.solver_->make_term(smt::Op(smt::Extract, integer - 1, 0), word->getTerm());
+                  }
+                  assert(res); //check res non-null
+                  $$ = new SMVnode(res,word_type);
+                }else{
+                  smt::PrimOp extendOp = word_type == SMVnode::Signed ? smt::Sign_Extend : smt::Zero_Extend;
+                  smt::Term res = enc.solver_->make_term(smt::Op(extendOp, integer - word_width), word->getTerm());
+                  assert(res); //check res non-null
```
CyanoKobalamyne: Same remark as for assert above.

frontends/smvparser.y
```diff
@@ -1271,7 +1369,7 @@ simple_expr: constant {
             | basic_expr OP_IN basic_expr {
                throw PonoException("No array");
             }
-            | basic_expr IF_ELSE basic_expr ":" basic_expr  {
+            | basic_expr IF_ELSE basic_expr ":" basic_expr %prec IF_ELSE {
```
CyanoKobalamyne, line 1382: What does this affect and why is it needed?

frontends/smvparser.y
```diff
@@ -1253,11 +1299,63 @@ simple_expr: constant {
...
+                  throw PonoException("Signed word conversion type is uncompatible");
+                }
+                assert(expr);
+                smt::Term res = enc.solver_->make_term(smt::Op(smt::Int_To_BV, $2), expr->getTerm());
+                assert(res);
+                $$ = new SMVnode(res, SMVnode::Signed);
+              }else{
+                SMVnode *integer = new constant(std::to_string($2));
+                $$ = new resize_expr($4,integer);
+              }
```
CyanoKobalamyne, line 1368: Where is this conversion coming from? As far as I can tell, the current NuSMV manual language description says that "signed" is used for unsigned-to-signed bit vector conversion. There is a separate `swconst` operator for converting an integer to a signed bit-vector constant.

tests/encoders/test_encoder_inputs.h
```diff
@@ -32,7 +32,11 @@ const unordered_map<string, pono::ProverResult> smv_inputs(
       { "simple_counter_integer_uf.smv", pono::ProverResult::FALSE },
       { "combined-false.smv", pono::ProverResult::FALSE },
       { "combined-true.smv", pono::ProverResult::TRUE },
+      { "counter_yosys.smv", pono::ProverResult::FALSE },
```
CyanoKobalamyne, line 35: Is there any particular reason for this ordering?

tests/encoders/inputs/smv/real_arithmetic.smv
```diff
@@ -0,0 +1,15 @@
...
+    _a : real; -- a
+    _b : unsigned word[3]; -- b
+  DEFINE
+    _$itor$arithmetic#sv#7$2_Y := toint(resize(_b, 3));
+    _c := _$itor$arithmetic#sv#7$2_Y + _a;
+    _$itor$arithmetic#sv#8$5_Y := toint(resize(signed(_b), 3));
+    _$add$arithmetic#sv#8$4_Y := _a + _$itor$arithmetic#sv#8$5_Y;
+    _$eq$arithmetic#sv#8$6_Y := resize(word1(_c = _$add$arithmetic#sv#8$4_Y), 1);
+    _$auto$rtlil#cc#2837#Not$9 := !resize(_$eq$arithmetic#sv#8$6_Y, 1);
+    _$auto$rtlil#cc#2884#And$11 := resize(_$auto$rtlil#cc#2837#Not$9, 1) & resize(0ub1_1, 1);
```
CyanoKobalamyne, line 13: What are these two variables?

tests/encoders/inputs/smv/real_arithmetic_floor.smv
```diff
@@ -0,0 +1,12 @@
...
+MODULE main
+  VAR
+    _VIN : real; -- VIN
+    _rst1 : unsigned word[1]; -- rst1
+    _rst2 : unsigned word[1]; -- rst2
+  DEFINE
+    _delta := 1.00000000000000000 / 8.00000000000000000;
+    _$div$maurice2024_adc#sv#46$14_Y := _VIN / _delta;
+    _$floor$maurice2024_adc#sv#46$15_Y := signed word[3](floor(_$div$maurice2024_adc#sv#46$14_Y));
+  INVARSPEC bool(word1(_$floor$maurice2024_adc#sv#46$15_Y = 0sb3_000));
```
CyanoKobalamyne, line 11: Shouldn't this hold? At least if we assume that the type conversion is mod 8.

CyanoKobalamyne (changes requested):
Hi! Thank you for contributing to Pono and sorry for taking so long to review your PR. I have left a couple of comments. :)

frontends/smv_node.cpp
augustomafra: The SMV parser actually does two passes: the first one generating an AST with SMVnodes whose only purpose is to be dumped as text with all modules flattened in SMVEncoder::preprocess(). Then the second pass reads the flattened stream and does the actual encoding to smt-switch terms. 

The first pass is run with enc.module_flat = false, and the second one with enc.module_flat = true, hence the branching on this condition almost everywhere in smvparser.y.

This flow is executed from SMVEncoder's constructor: https://github.com/stanford-centaur/pono/blob/309e07a3e6f7f20f3f793df27a7805ac6add9645/frontends/smv_encoder.h#L31-L38

Having that in mind, the issue here is that the flattened text had a syntax error: it would dump `resize a, 32` instead of `resize(a, 32)`

frontends/smvparser.y
augustomafra: It seems NuSMV/NuXmv manuals do not specify that, indeed. I figured out this rule by running the test tests/encoders/inputs/smv/ternary.smv with NuXmv directly, and it would accept it as right-associative whereas pono would raise a syntax error.

I hit this condition because some SMV files generated from Yosys contained right-associative ternaries, e.g. `result := a ? 0.0 : b ? 1.0 : c ? 2.0 : d ? 3.0 : 4.0;`

frontends/smvparser.y
augustomafra: Maybe there's a conceptual definition that I should double check regarding Pono's language frontend.
I see you mention NuSMV's manual (https://nusmv.fbk.eu/userman/v26/nusmv.pdf), however, I was actually following NuXmv's manual (https://nuxmv.fbk.eu/downloads/nuxmv-user-manual.pdf).

I assumed NuXmv was the reference used because real arithmetic was introduced only in it. Also, it introduced the signatures `integer * real → boolean` and `real * integer → boolean` for `=` and `!=`.

I don't know if Pono should stick strictly to SMV, as indeed there were no SMV tests with `real` types before my change. Let me know what is your decision here.
In any case, I would need to continue with the NuXmv style on my fork for the purpose of my work, so I could figure out later how to merge the related changes in SMV here.

frontends/smvparser.y
augustomafra: You are right, I checked that this constraint applies to nuXmv as well. I'm not sure why I needed Integers here, I need to confirm if any of my tests used it.

frontends/smvparser.y
augustomafra: This was another difference introduced in nuXmv. Again, it is fine if Pono would stick to SMV, then I could keep those changes limited to my fork.

frontends/smvparser.y
augustomafra: This complements the change making ternaries right-associative. The `%prec` directive tells Bison that the associativity of the entire rule follows the associative of the given token (`IF_ELSE`). The first change is not effective without this extra directive.

tests/encoders/inputs/smv/real_arithmetic.smv
augustomafra: For some reason (maybe a bug) Yosys always dumps those dangling variables at the end of the SMV definitions. They don't have any purpose, but I left them in the test only for ensuring that Yosys' output won't be rejected in the future.

tests/encoders/test_encoder_inputs.h
augustomafra: None at all, I only appended the new tests at the bottom. Please let me know if any other grouping is preferred (e.g. grouping by TRUE/FALSE results or alphabetical order)

tests/encoders/inputs/smv/real_arithmetic_floor.smv
augustomafra: This test is actually pretty underconstrained because `_VIN` is a free input. So any real value could propagate into the `floor` operation.
Maybe a more meaningful test case should have an actual true invariant to check for. My goal at first was just to ensure that the syntax and typing checks would pass.

frontends/smvparser.y
augustomafra: This was indeed wrong in my implementation. At first I encoded both `resize(a,N)` and `signed word[N](a)` using the same `SMVnode` type, so their signatures were merged in this code thus accepting both bitvectors and integers.

I have split them now into the corresponding `SMVnode` specializations.

augustomafra:
Hi, @CyanoKobalamyne ,

Sorry for the long time without updates. I'm resuming my work on the pull requests now and I'm addressing the comments you have raised.

Regarding the comments on the NuXmv's manual, I would like to check if adopting the specifications from the latest NuXmv version makes sense (https://nuxmv.fbk.eu/downloads/nuxmv-user-manual.pdf). I needed the NuXmv specific syntax in my work using Pono, so hopefully it fits for merging to the main repository as well. However, do let me know if you prefer that I keep this extension limited to my fork.

Thanks!

PR #476, opened 2025-09-29
Promote inputs in prop

### LINKED ISSUE #380: Unsoundness due to property monitor

Po-Chun-Chien:
Fixes #380

### DIFF
--- a/pono.cpp
+++ b/pono.cpp
@@ -35,6 +35,7 @@
 #include "printers/btor2_witness_printer.h"
 #include "printers/vcd_witness_printer.h"
 #include "smt-switch/logging_solver.h"
+#include "smt-switch/utils.h"
 #include "smt/available_solvers.h"
 #include "utils/logger.h"
 #include "utils/make_provers.h"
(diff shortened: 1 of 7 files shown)


pono.cpp
```diff
@@ -98,17 +99,24 @@ ProverResult check_prop(PonoOptions pono_options,
 
   if (pono_options.promote_inputvars_) {
     ts = promote_inputvars(ts);
-    assert(!ts.inputvars().size());
+  } else if (!ts.only_curr(prop) && ts.no_next(prop)) {
```
CyanoKobalamyne, line 102: We could also always promote input variables in the property, so the monitor would only contain current and next states.

ahmed-irfan:
what do you mean by "when the property has no next-state variables"

pono.cpp
```diff
@@ -98,17 +99,24 @@ ProverResult check_prop(PonoOptions pono_options,
...
+    UnorderedTermSet ivs_in_prop;
+    get_free_symbolic_consts(prop, ivs_in_prop);
+    ts = promote_inputvars(ts, ivs_in_prop);
   }
 
   bool has_monitor = false;
   if (!ts.only_curr(prop)) {
     logger.log(1,
-               "Got next state or input variables in property. "
+               "Got next state (and input variables) in property. "
```
CyanoKobalamyne:
```suggestion
               "Got next state in property. "
```

pono.cpp
Po-Chun-Chien: I’d prefer to make as few modifications to the TS as possible, since IIRC some engines distinguish between `ts.statevars()` and `ts.inputvars()`.

pono.cpp
Po-Chun-Chien: Applied in 0175c71

PR #482, opened 2025-10-06
Implement k-liveness

Po-Chun-Chien:


Po-Chun-Chien:
c4d2c1b was cherry-picked from 2d11cde in #488, such that the violation witnesses for justice can be printed

PR #510, opened 2026-01-01
Avoid excessive use of reset-assertions in interpolation-based engines

Po-Chun-Chien:
This PR replaces the use of reset-assertions with push/pop for adding and removing assertions before SAT checks in interpolation-based engines (interp, ismc, and dar). In addition, the fixed-point checks now become incremental.

I ran a quick evaluation with the following observations:

- When using `--smt-solver bzla`, performance improves with this change.
- When using `--smt-solver msat`, MathSAT sometimes performs better when assertions are reset.

Since Bitwuzla is our default solver, it makes sense to move forward with this change.

### DIFF
--- a/engines/dual_approx_reach.cpp
+++ b/engines/dual_approx_reach.cpp
@@ -426,16 +426,15 @@ bool DualApproxReach::check_fixed_point(const TermVec & reach_seq,
     solver_->pop();
     if (r.is_unsat()) {
       // reached a fixed point
-      solver_->pop(i);  // pop all assertions
+      solver_->pop();  // pop all assertions
       fixed_point = acc_img;
       return true;
     }
     // extend the accumulated reached set
-    solver_->push();
     solver_->assert_formula(solver_->make_term(Not, reach_seq.at(i)));
     acc_img = solver_->make_term(Or, acc_img, reach_seq.at(i));
   }
-  solver_->pop(reach_seq.size());  // pop all assertions
+  solver_->pop();  // pop all assertions
   return false;
 }
 
--- a/engines/interp_seq_mc.cpp
+++ b/engines/interp_seq_mc.cpp
@@ -263,16 +263,15 @@ bool InterpSeqMC::check_fixed_point()
     solver_->pop();
     if (r.is_unsat()) {
       // reached a fixed point
-      solver_->pop(i);  // pop all assertions
+      solver_->pop();  // pop all assertions
       invar_ = acc_img;
       return true;
     }
     // extend the accumulated reached set
-    solver_->push();
     solver_->assert_formula(solver_->make_term(Not, reach_seq_.at(i)));
     acc_img = solver_->make_term(Or, acc_img, reach_seq_.at(i));
   }
-  solver_->pop(reach_seq_.size());  // pop all assertions
+  solver_->pop();  // pop all assertions
   return false;
 }
 
--- a/engines/interpolantmc.cpp
+++ b/engines/interpolantmc.cpp
@@ -192,8 +192,8 @@ bool InterpolantMC::step(const int i)
       Ri = unroller_.at_time(unroller_.untime(Ri), 0);
 
       if (has_converged(Ri, R, interp_count)) {
-        assert(solver_->get_context_level() == interp_count);
-        solver_->pop(interp_count);
+        assert(solver_->get_context_level() == 1);
+        solver_->pop();
         logger.log(1, "Found a proof at bound: {}", i);
         invar_ = unroller_.untime(R);
         return true;
@@ -222,8 +222,8 @@ bool InterpolantMC::step(const int i)
       throw PonoException("Interpolant generation failed.");
     }
   }
-  assert(solver_->get_context_level() == interp_count + 1);
-  solver_->pop(interp_count + 1);
+  assert(solver_->get_context_level() == 1);
+  solver_->pop();
 
   // Note: important that it's for i > 0
   // transB can't have any symbols from time 0 in it
@@ -281,7 +281,7 @@ bool InterpolantMC::has_converged(const Term & new_itp,
     solver_->push();
     solver_->assert_formula(solver_->make_term(Not, reached));
   }
-  assert(solver_->get_context_level() == interp_count);
+  assert(solver_->get_context_level() == 1);
 
   // check if new_itp is already covered by the reached states
   solver_->push();
@@ -294,7 +294,6 @@ bool InterpolantMC::has_converged(const Term & new_itp,
   if (!r.is_unsat()) {
     // new_itp contains a state not in reached states
     // extend the reached states
-    solver_->push();
     solver_->assert_formula(solver_->make_term(Not, new_itp));
   }
   return r.is_unsat();


engines/dual_approx_reach.cpp
```diff
@@ -143,7 +143,7 @@ bool DualApproxReach::step(int i)
 
 bool DualApproxReach::step_0()
 {
-  solver_->reset_assertions();
+  solver_->pop(solver_->get_context_level());
```
ahmed-irfan: nit: may be it would be good to add a local method for this (call it soft_reset)

engines/dual_approx_reach.cpp
```diff
@@ -414,12 +406,24 @@ bool DualApproxReach::check_fixed_point(const TermVec & reach_seq,
...
+    solver_->push();
+    solver_->assert_formula(reach_seq.at(i));
+    Result r = solver_->check_sat();
+    assert(r.is_unsat() || r.is_sat());
+    if (r.is_unsat()) {
+      // reached a fixed point
       fixed_point = acc_img;
       return true;
     }
+    solver_->pop();
```
ahmed-irfan: do this on line 418 (before if block)

engines/interp_seq_mc.cpp
```diff
@@ -232,25 +240,30 @@ void InterpSeqMC::update_term_map(size_t i)
...
+    solver_->push();
+    solver_->assert_formula(reach_seq_.at(i));
+    Result r = solver_->check_sat();
+    assert(r.is_unsat() || r.is_sat());
+    if (r.is_unsat()) {
+      // reached a fixed point
       invar_ = acc_img;
       return true;
     }
+    solver_->pop();
```
ahmed-irfan: ditto -- move it before if block

engines/dual_approx_reach.cpp
CyanoKobalamyne: I don't think we need this line at all. There shouldn't be context levels above the current one when this function gets executed; if there are, and that interferes with solving, that's a user error.

engines/dual_approx_reach.cpp
```diff
@@ -414,12 +406,24 @@ bool DualApproxReach::check_fixed_point(const TermVec & reach_seq,
...
+  solver_->assert_formula(solver_->make_term(Not, acc_img));
   for (size_t i = 1; i < reach_seq.size(); ++i) {
-    if (check_entail(reach_seq.at(i), acc_img)) {
+    solver_->push();
+    solver_->assert_formula(reach_seq.at(i));
+    Result r = solver_->check_sat();
+    assert(r.is_unsat() || r.is_sat());
+    if (r.is_unsat()) {
+      // reached a fixed point
       fixed_point = acc_img;
```
CyanoKobalamyne, line 430: You should add a pop here to remove the assertion level introduced before the for loop.

engines/dual_approx_reach.cpp
```diff
@@ -414,12 +406,24 @@ bool DualApproxReach::check_fixed_point(const TermVec & reach_seq,
...
       return true;
     }
+    solver_->pop();
+    // extend the accumulated reached set
+    solver_->push();
```
CyanoKobalamyne: This push is unnecessary; there's no associated pop and there is already a new context level created before entering the loop.

engines/dual_approx_reach.cpp
CyanoKobalamyne: Agreed.

engines/interpolantmc.cpp
```diff
@@ -239,26 +237,56 @@ bool InterpolantMC::step(const int i)
...
-  solver_->reset_assertions();
-  solver_->assert_formula(
-      solver_->make_term(And, p, solver_->make_term(Not, q)));
+  if (solver_->get_context_level() == 0) {
+    // initialize the solver stack on the 1st call to this method
+    // at the current unrolling step
+    assert(reached == init0_);
+    solver_->push();
+    solver_->assert_formula(solver_->make_term(Not, reached));
+  }
```
CyanoKobalamyne: I don't think this code should be here; we should not be relying on the context level being 0 for terms that need to be asserted only once. Better to do it when we know we are in the first step.

engines/interp_seq_mc.cpp
Po-Chun-Chien: See 22e5476

engines/dual_approx_reach.cpp
Po-Chun-Chien: See 22e5476

engines/dual_approx_reach.cpp
Po-Chun-Chien: See 8b75922

engines/interpolantmc.cpp
```diff
@@ -239,26 +237,56 @@ bool InterpolantMC::step(const int i)
...
+  solver_->push();
+  solver_->assert_formula(new_itp);
+
   Result r = solver_->check_sat();
   assert(r.is_unsat() || r.is_sat());
+
+  if (!r.is_unsat()) {
+    // new_itp contains a state not in reached states
+    // extend the reached states
+    solver_->pop();
```
CyanoKobalamyne: Why is there no pop for the unsat branch?

engines/interpolantmc.cpp
```diff
@@ -183,23 +186,18 @@ bool InterpolantMC::step(const int i)
...
-      if (check_entail(Ri, R)) {
-        // check if the over-approximation has reached a fix-point
+      if (has_converged(Ri, R, interp_count)) {
         logger.log(1, "Found a proof at bound: {}", i);
         invar_ = unroller_.untime(R);
         return true;
-      } else {
-        logger.log(1, "Extending reached states (count: {})", interp_count);
-        logger.log(3, "Using interpolant: {}", Ri);
-        R = solver_->make_term(Or, R, Ri);
```
CyanoKobalamyne, line 194: I think it would probably be better to keep these here instead of moving them into the other function.

engines/interpolantmc.h
```diff
@@ -46,9 +46,33 @@ class InterpolantMC : public SafetyProver
...
-  bool check_entail(const smt::Term & p, const smt::Term & q);
+  /**
+   * @brief Check whether the reached states have converged to a fixed point,
+   * that is, whether the newly computed interpolant is already covered by the
+   * reached states.
+   *
+   * This method has the following side effects:
+   * - the solver stack is modified, and
+   * - the `reached` term is extended to include `new_itp` if `reached` does not
+   *   cover `new_itp` (i.e., the method returns `false`).
```
CyanoKobalamyne: It's not very intuitive that this happens in a function called `has_converged`. I think we should leave this side effect in the caller.

Po-Chun-Chien:
@CyanoKobalamyne I've reworked the implementation. Now push/pop operations are balanced during calls to `check_until()`/`prove()`, except when there is a CEX.

Note that we cannot pop all contexts immediately after `compute_witness()` in the engines because the CEGAR classes (e.g., `cegar_ops_uf`) may still rely on them. Changing that would require more effort beyond this PR.

CyanoKobalamyne (approved):

PR #513, opened 2026-01-15
Remove Flex header search from CMakeLists.txt

CyanoKobalamyne:
We aren't using them and they can't be found by default on, e.g., macOS.


CyanoKobalamyne:
`FlexLexer.h` is in fact included in the generated lexer source.

PR #520, opened 2026-01-30
Enforce use of relation TS when specified

Po-Chun-Chien:
This PR introduces the following changes:

- Add a new option `--use-rel-ts`
- Make `pseudo_init_and_prop()` accept only relation TS
- Create the correct type of TS right at the beginning (before encoding)

The latter two are important to avoid issues arising from casting based on `ts.is_functional()`.

In the original implementation (b19eafbe), `pseudo_init_and_prop()` may "convert" a functional TS to a relational one.
However, since we are assigning the returned `RelationalTransitionSystem` back to a `TransitionSystem&` in `pono.cpp`, object slicing could cause issues in subsequent casting.
For example, if we modify `pono.cpp` as follows:

```c
  // in check_prop
  if (pono_options.pseudo_init_prop_) {
    ts = pseudo_init_and_prop(ts, prop);
    if (!ts.is_functional()) {
      dynamic_cast<RelationalTransitionSystem &>(ts);
    }
  }
```

A `std::bad_cast` will be encountered (suppose `--pseudo-init-prop is enabled`).

### DIFF
--- a/modifiers/mod_ts_prop.cpp
+++ b/modifiers/mod_ts_prop.cpp
@@ -31,7 +31,7 @@ Term pseudo_init_and_prop(TransitionSystem & ts, const Term & prop)
 {
   assert(!ts.is_functional());
 
-  logger.log(1, "Modifying init and prop");
+  logger.log(1, "Creating pseudo-initial state and property monitor");
   RelationalTransitionSystem & rts =
       static_cast<RelationalTransitionSystem &>(ts);
   Sort boolsort = rts.make_sort(BOOL);


modifiers/mod_ts_prop.h
```diff
@@ -24,15 +24,13 @@ namespace pono {
...
  *
  *  @param ts the transition system to modify
- *  @param prop the property to modify
- *  @return the updated transition system
- *          the returned system is relational even if ts is functional
- *  Updates the prop in-place
+ *  @param prop the original property
+ *  @return the updated prop
  */
-TransitionSystem pseudo_init_and_prop(TransitionSystem & ts, smt::Term & prop);
```
Po-Chun-Chien, line 35: I also updated (and improved) the function signature of `pseudo_init_and_prop`. I'm also fine with reverting it for backward compatibility (though I’m not sure if anyone else is currently using this function).

That said, introducing a breaking change in a new version might also be acceptable.

modifiers/mod_ts_prop.cpp
```diff
@@ -27,28 +27,13 @@ using namespace std;
...
-      rts.add_inputvar(iv);
-    }
-    rts.set_init(ts.init());
-    rts.set_trans(ts.trans());
-  }
-
-  assert(!rts.is_functional());
+  assert(!ts.is_functional());
 
+  logger.log(1, "Modifying init and prop");
```
CyanoKobalamyne:
```suggestion
  logger.log(1, "Creating pseudo-initial state and property monitor");
```

modifiers/mod_ts_prop.h
CyanoKobalamyne: I think this is fine, we have still not released a proper 2.0 version.

Po-Chun-Chien:
> I agree with the pseudo init fixes and changes, but what is the use case for the use-rel option?

I added it because I came across an interesting [case](https://github.com/stanford-centaur/pono/pull/521#issuecomment-3824313833) where it can help improve performance when combined with the transformation introduced in #521.

That said, I haven’t evaluated it on a larger benchmark set, so I’m not certain of its overall impact.
I'm also fine with not exposing this option to users.

PR #533, opened 2026-05-06
SystemVerilog frontend

CyanoKobalamyne:


CyanoKobalamyne:
Claude's self-review of the current state of this code, for future reference:

Overall verdict: **high-quality, well-documented branch**, but with two concrete bugs and one build config issue worth fixing before merge.

**Comments/docs** — Standout quality, well above the codebase's norm (non-empty `\brief`s, design-decision rationale documented). One stale comment at `systemverilog_encoder.cpp:3248` claims dynamic array indexing is restricted to single-bit elements, but the code below actually handles arbitrary element widths correctly — just needs correcting/removing. Minor: both new files list "Top contributors: None" instead of actual authors.

**Architecture** — Fits the existing BTOR2/SMV/VMT frontend pattern well; slang types are cleanly forward-declared out of the header; the ~4800-line `.cpp` is large but organized into clear, banner-commented sections rather than being genuinely disorganized (splitting SVA handling into its own file would be a nice follow-up, not a blocker). State ownership (symbol/wire/alias maps, pre-scan ordering) is deliberate and handles real hazards like forward references.

**Bugs found:**
1. **Critical, undocumented soundness gap**: signed SystemVerilog types (`signed`, `integer`) are silently treated as unsigned everywhere — zero-extension only, no `BVSlt`/`BVSdiv`/sign-extend path anywhere in `expr_to_term`/`resize_to`. No test coverage exists for signed types, and it's not in the "known unsupported" ledger, so a design using signed arithmetic could produce a wrong verdict with no warning. Recommended fix: either implement signed-aware ops, or explicitly throw for signed operands (consistent with how the file handles other unsupported constructs elsewhere).
2. **Real bug, untested path**: in `process_continuous_assign` (`systemverilog_encoder.cpp:1329-1339`), the "is this a full write" check for a wire's *first* write only checks `lo == 0`, not `hi` against the symbol's true declared width — unlike the correct parallel logic in `commit_write`. This is reachable when an instance-array element writes a slice of a shared parent bus (documented as a supported feature) and can produce out-of-range `Extract` calls or a corrupted term. No test currently exercises this scenario.
3. **Build config**: `CMakeLists.txt:386-390` sets `CXX_STANDARD 20` on the whole `pono-lib` target when `--with-slang` is enabled, silently recompiling all existing C++17 sources as C++20. Should be scoped to just `systemverilog_encoder.cpp` via `set_source_files_properties`.
4. Noted in passing (pre-existing, unrelated to this branch): a missing `break;` in `options.cpp:972-974` causes `--no-ic3ia-reduce-preds` to also disable two unrelated options via fallthrough.

Everything else — bit-width/range math, casex/casez and wildcard-equality handling, loop-unrolling caps, and the SVA/LTL tableau construction — checked out clean and was called out as well-engineered.

PR #535, opened 2026-05-26
Mention new FM 2026 paper and HWMCC 2025 award

Po-Chun-Chien:
Also, replace paper URLs with DOI links


CyanoKobalamyne (approved):
Thank you! :)

